PR #21, opened 2015-06-16
[READY] NPM Module

SethNute:
screenshot from 2015-06-18 11 46 17

Above are the files that will be included in the node module via <code>npm install canon-react</code>, dist contains the min version of canon-react and demo contains all the files needed for index.html to display properly (using components defined in canon-react.min.js). 

How to use:
Install via <code>npm install canon-react</code>
Where you would like to use react components from canon-react, add <code>var componentName = require('canon-react').componentName;</code>

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "description": "Rackspace Canon components built with React",
   "scripts": {
+    "prepublish": "./prepublish.sh",
     "build": "./node_modules/.bin/grunt build",
     "demo-build": "./node_modules/.bin/grunt demo-build",
     "test-watch": "./node_modules/.bin/grunt watch 2>&1 >/dev/null & ./node_modules/karma/bin/karma start karma.dev.js",
--- a/package.json
+++ b/package.json
@@ -2,6 +2,10 @@
   "name": "canon-react",
   "version": "0.1.0",
   "description": "Rackspace Canon components built with React",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/rackerlabs/canon-react"
+  },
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -2,7 +2,12 @@
   "name": "canon-react",
   "version": "0.1.0",
   "description": "Rackspace Canon components built with React",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/rackerlabs/canon-react"
+  },
```
SethNute: Rackerlabs is all open source stuff, right?

bradgignac: There is also some private stuff.

src/Button.jsx
```diff
@@ -1,3 +1,5 @@
+var React = require('react');
```
SethNute: I understand that here node looks for the package react in a few different places (global and local modules). How was the symbol React being provided globally before adding these require statements?

package.json
SethNute: So is this okay to include or no?

SethNute:
Can I go ahead and <code>npm publish</code> for testing purposes? I've tested this by installing it globally on my local machine and requiring it from the node terminal so far.

package.json
adampickeral: the private things are actually in private repos, so this should be fine. @bradgignac can correct me if i'm wrong

bradgignac: Oh yeah. It's fine.

adampickeral:
looks pretty good to me, give it a shot!

adampickeral:
:+1: :exclamation:

adampickeral:
:shipit:

PR #24, opened 2015-06-19
Add src to npm module

SethNute:


bradgignac:
It *seems* like the default would be to import the raw source files rather than a built and minified version. If you're using CJS, you're either using a build system (browserify, webpack) or running on the server (node, iojs). If you need a built and minified version, you probably aren't using CJS and won't be able to simply `require('canon-bootstrap)` anyway and will have to manually link to the built file.

bradgignac:
And by import, I mean point `main` in package.json to an unbuilt file.

SethNute:
Makes sense, I was still in the mind set that I didn't have access to the raw src for whatever reason. This look good now?

bradgignac:
Works for me. Will this work in Reach's build?

/cc @adampickeral

adampickeral:
yeah this looks good to me. Reach is using a built version currently that we're just grabbing from the github release.

adampickeral:
add `external: ['react']` under the options to `browserify: release` in Gruntfile to make this work with reach

adampickeral:
specifying `external: ['react']` doesn't quite make it work in reach. we need to be able to have all of the `var React = require('react');` statements removed from the source when building it. i tried digging around a bit the other day but couldn't quite get it to work. so as it stands right now the current version won't work with reach.

bradgignac:
@adampickeral https://github.com/thlorenz/browserify-shim

adampickeral:
thanks @bradgignac that did the trick. @SethNute you can cherry-pick this commit to make that fix https://github.com/rackerlabs/canon-react/commit/0be8800296e572d1a9181d879ed1a20ace3fdfc7

package.json
```diff
@@ -17,13 +17,23 @@
...
+    ]
+  },
+  "browserify-shim": {
+    "react": "global:React"
+  },
   "peerDependencies": {
     "react": ">=0.13.2"
   },
   "devDependencies": {
+    "browserify": "<4.0.0",
```
SethNute: Our version of grunt-browserify requires this be less than 4

SethNute:
@adampickeral If this looks good to you I'm gonna go ahead and publish it to make sure it installs properly with <code>npm install canon-react</code>

src/Button.jsx
```diff
@@ -1,5 +1,3 @@
-var React = require('react');
```
adampickeral: you still want to leave these

adampickeral: the browserify-shim will change this to something like `var React = (window.React)` in the bundled version. that's the version that will work with non-commonJS systems (like control panel). systems using commonjs like modules will still need `var React = require('react')` in the source files.

package.json
adampickeral: :+1:

SethNute:
@adampickeral This look good to go now? Just threw those require statements back in.

PR #26, opened 2015-06-19
Eslint

underbyte:


.eslintrc
```diff
@@ -0,0 +1,47 @@
...
+    "jasmine": true
+  },
+  "plugins": [
+    "react"
+  ],
+  "rules": {
+    "semi": 2,
+    "curly": 2,
+    "strict": 0,
+    "quotes": 0,
```
willcumbie: IMO we should enforce single-quotes, while we still can. I would have done this on Reach if I could've, but fixing all the errors in a codebase that big would've taken weeks.

.eslintrc
```diff
@@ -0,0 +1,47 @@
...
+    "strict": 0,
+    "no-underscore-dangle": 0,
+    "camelcase": 0,
+    "dot-notation": 0,
```
willcumbie: Surprised we can't use this. We need bracket notation in Reach because of the Closure compiler--we don't need it here, do we?

package.json
```diff
@@ -49,9 +49,12 @@
   },
   "devDependencies": {
     "babel": "^5.5.8",
+    "babel-runtime": "^5.5.8",
     "babelify": "^6.1.2",
     "browserify": "<4.0.0",
     "browserify-shim": "3.8.9",
+    "eslint": "^0.23.0",
+    "eslint-plugin-react": "^2.5.2",
```
willcumbie: +1 to using the React ESLint plugin.

willcumbie:
:ship:

adampickeral:
yes :sparkler: :shipit:

IanVS:
FYI, ESLint released `1.x`, and there is a new eslint-plugin-react major version to match.

underbyte:
@IanVS thank you! I'll look into it before merging

PR #27, opened 2015-07-01
Popover

adampickeral:
image

http://rackerlabs.github.io/canon/ui-components/#popover

## Components
`Popover`
`PopoverOverlay`
`PopoverBody`
`PopoverFooter`

### `Popover`
The `Popover` component controls the behavior of the popover itself (positioning, open/closed, request for close, etc.) The `Popover` has four properties:
* `placement`
* `isOpen`
* `onRequestClose`
* `target`

###### `placement`
`placement` defines the position of the popover in relation to its target. Values are one of `'right', 'bottom-right', 'left', 'bottom-left'

###### `isOpen`
`isOpen` accepts boolean values and defines whether or not the popover is open or closed. It is up to the parent component to decide whether or not a popover should be displayed. This is necessary to allow for such things as allowing the processing of a command to finish before closing a popover.

###### `onRequestClose`
`onRequestClose` is a callback that is used to notify a parent component that some action has occured requesting to close a popover. This could be the clicking of a cancel link on the popover. Events that the popover itself monitors and invokes the callback include the pressing of the Escape key and clicking off of the popover.

###### `target`
`target` is the id of the element at which the popover will be pointing.

### `PopoverOverlay`
The `PopoverOverlay` class renders the outer markup of a popover, as well as the arrow of the popover. It accepts a `placement` prop (passed to it from the `Popover`) which is used to determine the position of the arrow relative to the target at  which the popover will be pointing. The values accepted are the same as those of the `PopoverTrigger`.

### `PopoverBody` and `PopoverFooter`
These components should be used together when displaying a popover that will do some sort of form processing. If a popover is only being used to display information, these components aren't needed. The body and footer components simple wrap their children in the appropriate markup for a popover form. They don't do any special processing of the children to be rendered. Any change of state of buttons rendered in the footer is the responsibility of the consumer to update. This allows for greater flexibility of what the body and footer can be composed of.

###### Render the Trigger and the Popover
```javascript
  // Parent component rendering the popover and the trigger
  {
    showUpdatePopover: function () {
      this.setState({shouldShowUpdateNamePopover: true})
    },

    requestCloseCallback: function () {
      this.setState({
        shouldShowUpdateNamePopover: this.getState().isUpdateNameProcessing
      });
    },

    render: function () {
      return (
        <div>  
          <Button id='target-element-id' onClick={this.showUpdatePopover}>Hello</Button>
          <UpdateNamePopover placement='right' isOpen={this.getState().shouldShowUpdateNamePopover} onRequestClose={this.requestCloseCallback} target='target-element-id' />
        </div>
      );
    }
  }
```

###### Popover
```javascript
var UpdateNamePopover = React.createClass({
  render: function () {
    return (
      <Popover placement={this.props.placement} isOpen={this.props.isOpen} onRequestClose={this.props.onRequestClose} target={this.props.target}>
        <PopoverOverlay>
          <PopoverBody>
            <form>
             <label>Name</label>
             <input type='text' />
            </form>
          </PopoverBody>
          <PopoverFooter>
            <Button type='primary' onClick={this._processSave}>Save</Button>
            <Button type='link' onClick={this.props.onRequestClose}>Cancel</Button>
            <ProcessingIndicator hidden={true} />
          </PopoverFooter>
        </PopoverOverlay>
      </Popover>
    );
  },

  _processSave: function () {
    // set some state to toggle button states for processing of action
    // perform some action, maybe through stores
    // on successful action completion, invoke this.props.onRequestClose.
    // OR, register for events in a store and update state appropriately
  }
});
```


src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,178 @@
...
+  },
+
+  _renderPopover: function () {
+    var popover;
+
+    popover = React.cloneElement(
+      this.props.popover,
+      {
+        placement: this.props.placement,
+        hideCallback: this._hide
```
jdrush89: Is the hideCallback doing anything right now?  Should hideCallback be in the propTypes of the Popover class, and that class should have a public hide function that triggers the callback?

src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,178 @@
...
+  },
+
+  _handleDocumentClick: function (e) {
+    if (React.findDOMNode(this._popoverNode).contains(e.target)) {
+      return;
+    }
+    this._hide();
+  },
+
+  _handleEscape: function () {
```
jdrush89: _handleEscape and _handleEscapePress are hard to distinguish from their names, maybe change to _handleEscape _listenToEscape?  (same for _handleClicksOutsideOfPopover and _handleRootCloseEvents)

demo/demo.jsx
```diff
@@ -6,12 +6,49 @@ var Button = CanonReact.Button;
...
 // var ProcessingIndicator = require('./ProcessingIndicator');
 // var ProgressBar = require('./ProgressBar');
+// var Popover = require('./Popover');
+// var PopoverBody = require('./PopoverBody');
+// var PopoverFooter = require('./PopoverFooter');
+// var PopoverTrigger = require('./PopoverTrigger');
+// var ProcessingIndicator = require('./ProcessingIndicator');
+
+var DemoPopover = React.createClass({
+  propType: {
```
jdrush89: Should the propTypes for this class be 
```propType {
  placement: React.PropTypes.oneOf(['right', 'bottom-right', 'left', 'bottom-left'])
  hideCallback: React.PropTypes.func
}```

**Comment by bradgignac** [comment]
Do you need a transparent overlay to handle clicking outside the popover?

**Comment by bradgignac** [comment]
Also, if you don't move this to its own rendering tree underneath the body, I think it will be clipped in certain layouts.

**Comment by adampickeral** [comment]
@bradgignac clicking outside of the popover is determined by checking if the target of the click is contained in the popover itself https://github.com/rackerlabs/canon-react/pull/27/files#diff-2f304169bd79349a64a2f345081b689dR161

i think i need some more explanation or examples to understand your second comment about a rendering tree underneath the body

**Comment by bradgignac** [comment]
@adampickeral IIRC, the intention of the overlay is to prevent interaction with elements outside the popover. For example, you shouldn't trigger tooltips on elements outside the popover. Not sure how important this behavior is, but I think this is the existing behavior.

For the second comment, render a popover inside a 10x10 box with `overflow: hidden` set. It will cause the popover the be clipped to the containing box. By rendering the popover in a second tree directly underneath the body, it will never be clipped. For reference, most of the Closure pop-ups (tooltips, dropdowns, modals, etc.) do this. I think Tether handles this for you in some configurations (or maybe a derivative library), but I'm not sure how that plays with DOM diffing.

**Comment by bradgignac** [comment]
Never mind, looks like you're already rendering to a separate tree.

**Comment by bradgignac** [comment]
One more question: under the current design, how will you prevent a popover from closing while an action is progress?

**Comment by adampickeral** [comment]
@bradgignac thanks for the feedback on the overlay to prevent other actions. i think that is a good behavior to keep, so i'll incorporate that into the design.

as far as preventing the popover from closing while an action is in progress, i hadn't even thought about that yet. i'll work that in too

**Comment by bradgignac** [comment]
@adampickeral IIRC, I've seen modals handle this with two properties on the popover:

1. `isOpen` toggles popover visibility.
2. `onRequestClose` calls back to notify that the user attempted to close the popover. It's up to the owner of the popover to decide if that's allowed and toggle `isOpen` accordingly.

At that point, it would be pretty easy to wire this up to a higher-order `Command` component in an application.

File: src/ButtonGroup.jsx [hunk 4]
```diff
@@ -2,8 +2,14 @@ var React = require('react');
 
 var ButtonGroup = React.createClass({
   render: function () {
+    var classes;
+
+    classes = [];
```
willcumbie: Why not just `classes = [this.props.className, 'rs-btn-group'];`?

src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+  _getTetherConfig: function () {
+    var tetherConfig;
+
+    switch (this.props.placement) {
+      case 'left':
+        tetherConfig = {
+          attachment: 'top right',
+          targetAttachment: 'middle left',
+          offset: '38px 20px'
```
willcumbie: Where are these offsets coming from? Trial and error?

willcumbie:
What do you do if you need to show a popover based on logic, not based on a click? For instance, the pattern in some parts of Reach where submitting one popover causes another to open.

bradgignac:
Based on @willcumbie's comments, I'm wondering if the actual tethering logic belongs inside the popover, and a target is specified as a property of the popover. `PopoverTrigger` then becomes just a convenience wrapper for click-based manipulation of the `isOpen` property of the popover.

bradgignac:
Also, moving tethering logic into the popover rather than the popover trigger makes it easier to assign a URL in which a popover is already open.

src/PopoverTrigger.jsx
```diff
@@ -0,0 +1,175 @@
...
+  componentDidMount: function () {
+    this._containerDiv = document.createElement('div');
+    document.body.appendChild(this._containerDiv);
+  },
+
+  render: function () {
+    var props;
+
+    props = {};
+    props.onClick = this._onTriggerClick;
```
j00k: this won't let you open a popover by for example selecting a button with the keyboard and pressing enter.

adampickeral: yes it will

src/PopoverTrigger.jsx
adampickeral: a little bit of trial and error and a little bit of looking at the position and size properties of the popover arrow

src/PopoverTrigger.jsx
j00k: TIL native buttons actually do this

adampickeral:
@bradgignac @willcumbie @j00k @jdrush89 ready for more feedback! I've updated the description of the PR to include the new design. The popover component now controls the tethering, and whether or not the popover is displayed is a function of the `isOpen` property on the popover component. The popover now also has a `target` property which accepts and id of the element it will be pointing to.

src/Popover.jsx
```diff
@@ -0,0 +1,161 @@
+var React = require('react');
+var Tether = require('tether');
+var PopoverBackground = require('./PopoverBackground');
+
+var Popover = React.createClass({
+
+  propTypes: {
```
willcumbie: Should include `target`.

src/Popover.jsx
```diff
@@ -0,0 +1,161 @@
...
+      default:
+        tetherConfig = {
+          attachment: 'top left',
+          targetAttachment: 'middle right',
+          offset: '38px -20px'
+        };
+    }
+
+    tetherConfig.element = React.findDOMNode(this._containerDiv);
+    tetherConfig.target = document.getElementById(this.props.target);
```
willcumbie: I'm not crazy about expecting an element ID for `target`. What's the advantage of this over just passing an element?

src/Popover.jsx
```diff
@@ -0,0 +1,161 @@
...
+
+  _getTetherConfig: function () {
+    var tetherConfig;
+
+    switch (this.props.placement) {
+      case 'left':
+        tetherConfig = {
+          attachment: 'top right',
+          targetAttachment: 'middle left',
+          offset: '38px 20px'
```
willcumbie: Are these offsets just based on what subjectively "looks right"? Would a CSS stylesheet be a more appropriate place for that?

src/Popover.jsx
```diff
@@ -0,0 +1,161 @@
...
+        placement: this.props.placement
+      }
+    );
+    this._popoverNode = React.render(popover, this._containerDiv);
+    this._tether = this._createTether(this._getTetherConfig());
+  },
+
+  // This is a seam for testing
+  _createTether: function (tetherConfig) {
+    return new Tether(tetherConfig);
```
willcumbie: +1 to using an external library to handle this.

src/Popover.jsx
adampickeral: when you actually define the popover to be rendered and the target at which it will be pointing, the elements don't exist yet. i couldn't think of way to design around that

src/Popover.jsx
adampickeral: it's sort of based on subjectively what looks good and roughly what the positioning of the arrow is. i might have to look at our reach code to figure out how we get that positioning just right

src/Popover.jsx
willcumbie: That's fair. It'd probably be possible to handle it using refs or something, but this is more straightforward/flexible.

src/Popover.jsx
adampickeral: it looks like we use some pretty magic numbers in reach for these kinds of offsets https://github.com/racker/reach/blob/master/cloudkick/webapp/site_media/js/ck/widgets/popover.js#L114

i'm not sure where CSS for this would live actually. do we bundle it with canon-react? i think these offsets might be tied to the logic you're using to do the actual positioning, so it might not make sense to live in something more general, like canon.

I'm into new waters here when it comes to positioning and figuring out the right placement (even though tether is handling most of the hard work), so I'm certainly open to suggestions

willcumbie: I think it would make sense to begin including custom CSS in canon-react, to handle cases like this. The big issue with setting the arrow positioning with inline styling, that I can think of anyway, is that it becomes impossible for the user to override it if they want something different. I don't know if it's standard for React libraries to use a lot of inline styling, but it seems undesirable to me.

adampickeral: so one thing we could do is allow users to specify the offset. that's a value that gets passed to tether, so we could expose that here and let users override what we're doing if they want to

adampickeral: also to be clear, the positioning of the arrow is controlled by canon css. it's the offset of the popover that's being changed here, and that's relative to the arrow element

willcumbie: Optionally overriding the offset makes sense to me; it'd be less trouble than packaging extra CSS (although I think it's likely we'll have to do that eventually sooner or later).

src/Popover.jsx
jdrush89: How would this work with tables where each row can be the target of a popover?  Each table row isn't going to have a different id right?

jdrush89:
I liked the idea of having the hide/show logic in a separate component (mostly because the trigger used for the dropdown was nearly identical).  I don't think the first solution has to be a perfect one, so if you want to roll with the logic in the popover that's fine.  What do you think is the best way to organize these components in the future though, do you think it would make sense to pull the isOpen prop into the popover trigger so it becomes a component that shows another component when a target is clicked or a prop is passed in?

src/Popover.jsx
adampickeral: good point @jdrush89 . i'm still not sure how to pass the target in initially if it hasn't been rendered yet and you want the popover to be displayed up front. maybe it involves using refs like @willcumbie suggested.

if everyone is ok with this implementation for the first pass then we can move forward with it and work on a better argument for the target when we need it.

adampickeral:
@jdrush89 if the state of a popover being displayed was simply a matter of clicking on a trigger then a popover trigger would make sense, but since the display of the popover is also tied to the execution of the command, it becomes more difficult to reconcile the state of the popover being open with the props in the trigger that determine if it is open.

for something simpler, however, such as menu dropdowns, a tigger might be more appropriate. we can likely pull the logic for showing/hiding/positioning into a mixin (or higher level class now that we have babel for es6!)

jdrush89:
@adampickeral sounds good. :ship:

adampickeral:
here's a pretty rudimentary example of using the popover in reach https://github.com/racker/reach/pull/13729 . this isn't tying into the command yet, and the open/closed logic is hacked together just to get it to display, but it wouldn't be hard to fully implement this.

i do think this already shows that it might not be ideal for `target` to be an ID though. this existing popover in this example was pointing at a label, which doesn't really need a special identifier. i'll see if i can figure out what the pattern would look like for passing an element as the `target`

willcumbie:
@adampickeral Yeah, I'm a bit worried by the possibility of the IDs going out of sync (or duplicate IDs appearing as we reuse classes). What about passing in a function which returns an element? That function could do the `refs` logic, or query by class or ID (or just return the parent), depending on what's needed.

src/Popover.jsx
j00k: :+1: we can work around these other cases as necessary

j00k:
:stars:

bradgignac:
+1

adampickeral:
i'm going to merge this in as is to get some more usage feedback, then work on that feedback as well as alternatives to target being an id in some fast follows

PR #33, opened 2015-08-21
Popover improvements

adampickeral:
Changes in this PR:

1. All users to override the offset of a popover
2. `target` can be a string ID or a function that returns the target element
3. `center` placement property. This will place the popover over the center of the target with no arrow rendered. This will be useful for modal like behavior where the popover should be placed over the center of the document body.


adampickeral:
@willcumbie @underbyte @j00k reviews please

willcumbie:
LGTM

bradgignac:
Ping @atarola.

underbyte:
:+1:

PR #36, opened 2015-09-02
Tooltip

adampickeral:
```jsx
let tooltipContent = // any react node (string, array, component, etc.);
<TooltipTrigger placement='top' content={tooltipContent}>
  <span>Hello</span>
</TooltipTrigger>
```

The `TooltipTrigger` renders the specified trigger (it's child) and a `Tooltip` component that will point to the trigger as the target. The tooltip will be rendered on hover of the target, and will be rendered to the bottom right of the target. Tether settings are configured so the tooltip will be rendered in the visible portion of the window (meaning it'll flip it's position if the target is too close to the right or bottom of the page).


willcumbie:
Does displaying on the bottom-right mean that link cell tooltips are going to look funny now, since the hover target for them is the entire cell (including all the whitespace)?

adampickeral:
i don't think you would make the entire cell the target, just the content in the cell

adampickeral:
also that isn't the current behavior anyways, the target is the link, not the whole cell

willcumbie:
lb-table-hover
I don't have a problem with changing it, though.

demo/DemoTooltipSection.jsx
```diff
@@ -0,0 +1,23 @@
+import Button from './Button';
+import React from 'react';
+import TooltipTrigger from './TooltipTrigger';
+
+class DemoTooltipSection extends React.Component {
+  render() {
+    const tooltipContent = (<span>This is a tooltip</span>);
+    return (
+      <div className='rs-detail-section'>
+        <div className='rs-detail-section-header'>
```
underbyte: we need a detail section and detail section header component

adampickeral:
weird, server list doesn't do that. seems kind of weird that the link takes up the entire cell like that

adampickeral:
@willcumbie i could make the placement a configurable option

underbyte:
looks good

willcumbie:
@adampickeral Yeah, server list is an odd case because it can have multiple tooltips (based on the icons). A configurable option would be cool.

src/Tooltip.jsx
```diff
@@ -0,0 +1,88 @@
...
+    if (this._tooltipNode) {
+      React.unmountComponentAtNode(this._containerDiv);
+      this._tooltipNode = null;
+    }
+  }
+
+  _show() {
+    this._containerDiv.className += ' visible';
+    this._tooltipNode = React.render(
+      <TooltipInnerContent onMouseOver={this.props.onMouseOver} onMouseLeave={this.props.onMouseLeave}>
```
willcumbie: Do we really need an entire component definition to encapsulate a single className?

adampickeral:
@willcumbie made the adjustments to allow configuration of tooltip placement and removed inner tooltip content component

src/TooltipTrigger.jsx
```diff
@@ -0,0 +1,84 @@
...
+      </trigger>
+    );
+  }
+
+  _showTooltip() {
+    setTimeout(() => {this.setState({isTooltipOpen: true}); }, 200);
+  }
+
+  _hideTooltip() {
+    setTimeout(() => {this.setState({isTooltipOpen: false}); }, 200);
```
jdrush89: If you quickly hover on and off the target, will the tooltip show up and disappear?  You might need to use timers so you can clear the interval when the mouse leaves the target.

src/TooltipTrigger.jsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  _shouldShowTooltip() {
+    return this.state.isTooltipOpen || this.state.isMouseInTooltip;
```
jdrush89: As soon as the mouse leaves the tooltip, it disappears?  Our current behavior is to reset the hide timer.  Not super important, but I think it makes copying info off a tooltip easier (don't have to worry about it disappearing if you accidentally moved the mouse a little too far).

src/TooltipTrigger.jsx
```diff
@@ -0,0 +1,84 @@
...
+    this.state = {
+      isTooltipOpen: false,
+      isMouseInTooltip: false
+    };
+  }
+
+  render() {
+    let triggerProps, showTooltipfunc, hideTooltipFunc;
+
+    showTooltipfunc = () => { this._showTooltip(); };
```
willcumbie: What's the difference between doing this and just doing `this._showTooltip.bind(this)`? I feel like doing `bind()` would be more explicit.

src/TooltipTrigger.jsx
```diff
@@ -0,0 +1,84 @@
...
+      onMouseLeave: hideTooltipFunc,
+      onFocus: showTooltipfunc,
+      onBlur: hideTooltipFunc,
+      ref: 'trigger'
+    };
+
+    this._trigger = React.cloneElement(React.Children.only(this.props.children), triggerProps);
+
+    return (
+      <trigger>
```
willcumbie: Where is `<trigger>` getting defined? That's not standard HTML, is it?

src/TooltipTrigger.jsx
adampickeral: learn you some es6.

arrow notation automatically binds `this` to the enclosing context

willcumbie: `bind` also does that, but it's more explicit. That's what I was saying.

src/TooltipTrigger.jsx
adampickeral: it's not standard. it's needed because react components have to return a single component in their render methods, i.e. you can't return fragments like this:
```
<Button />
<Tooltip />
```

I chose to use a non standard HTML element to prevent unintended layout side effects that could  be caused by adding a `<div>` or some other element to the dom. I did test this out and saw these unintended effects when using a `<div>` to wrap the trigger.

src/TooltipTrigger.jsx
adampickeral: it is and it isn't. arrow notation is designed to eliminate a lot of the confusion with `this` in js and I think we should leverage that when we have the ability to use and become more familiar with newer es6 features.

src/TooltipTrigger.jsx
willcumbie: Ok. I don't like it, but it does seem like a reasonable approach until React removes that limitation (if it ever does).

adampickeral: i agree, i don't really like it either. i'm not sure if it's something they will fix. a couple of issues i have seen on github have responses that it's a really difficult problem due to the way react works internally. we'll see i suppose

src/TooltipTrigger.jsx
willcumbie: :neutral_face: The purpose of the line of code is to bind `this`. Arrow notation isn't eliminating any confusion here, and I'm not a fan of using any language construct for its own sake. But I don't want to quibble over this, if you prefer the syntax, go for it.

adampickeral: i see your point. i have some changes to make to incorporate setting and clearing timers, so i'll clean this up if this code is still here after that.

willcumbie:
Code looks good to me, but the demo page doesn't seem to work--the tooltip never disappears.

src/TooltipTrigger.jsx
jdrush89: You don't need to create new functions right?  React automatically binds to this when you reference component functions, you can just do onMouseOver: this._showTooltip

jdrush89: It actually yells at you if you bind to this:
"You are binding a component method to the component. React does this for you automatically in a high-performance way, so you can safely remove this call."

adampickeral: that seems to change when you're using classes. it could be something with the way the code is transpiled down to es5, but autobinding doesn't happen anymore.

willcumbie: React doesn't autobind for ES6 classes.

I'm pretty sure it only autobinds certain functions in ES5, also. At one point it wasn't autobinding `forceUpdate`. This may have been fixed in a recent version, though.

adampickeral:
@willcumbie would you mind trying again? i haven't had any problems with the tooltip not disappearing

willcumbie:
Got it working now.

cooper-thumbs-up

j00k:
:ship:

PR #37, opened 2015-09-07
fix eslint warnings

underbyte:
- added eslint to package.json
- fixed all eslint warnings

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,7 +25,7 @@
     "radix": 2,
     "space-before-blocks": 2,
     "no-multiple-empty-lines": 2,
-    "one-var": 2,
+    "one-var": [2, { const: "never" }],
     "brace-style": [ 2, "1tbs", { "allowSingleLine": true }],
     "react/jsx-uses-vars": 1,
   },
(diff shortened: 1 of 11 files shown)


src/ProcessingIndicator.jsx
```diff
@@ -18,6 +18,7 @@ class ProcessingIndicator extends React.Component {
 }
 
 ProcessingIndicator.propTypes = {
+  className: React.PropTypes.string,
```
adampickeral: not a prop

.eslintrc
```diff
@@ -25,7 +25,7 @@
     "radix": 2,
     "space-before-blocks": 2,
     "no-multiple-empty-lines": 2,
-    "one-var": 2,
+    "one-var": [2, { const: "never" }],
```
adampickeral: what does this do?

underbyte: use one var/let/const declaration with assignment, the `never` ignores that rule for `const`

adampickeral: i don't think i fully understand, does that mean you don't have multiple lets or something? what is an example in which this would break?

Issue #15, opened 2015-06-09
Action Dropdown


adampickeral:
http://rackerlabs.github.io/canon/ui-components/#dropdowns

jdrush89:
I started working on this last hackday, will likely continue for hack week

Jigarsolanki:
@jdrush89 Could you please update this issue briefly with your approach?

jdrush89:
The ActionMenu component will create a container on the document in which renders an ActionMenuList into.  The ActionMenu passes its children to the ActionMenuList (the list items).  In addition, it renders a button whose content is set through props (could be a cog menu, action button, etc).  The menu list takes the button element as a target prop which it uses to do positioning.  The list items take as props events to fire when clicked.

bradgignac:
@jdrush89 How do you plan to handle position inside elements with `overflow: hidden`? Typically, we've had to render these components into the body rather than directly in the DOM. Similarly, how do you plan to handle cases in which the dropdown need to be displayed above the button?

jdrush89:
the actionMenuList will be rendered on a container on the document in all cases, similar to the tooltip we use today https://github.com/racker/reach/blob/master/cloudkick/webapp/site_media/js/rs/tooltip/tooltip.jsx#L88.  It uses absolute positioning based off the target's position.  First implementation will probably be dumb and always display under the target, but it shouldn't be too difficult to detect if the menu won't fit below and display it above instead in componentDidUpdate, similar to the repositioning I did for tooltips in this PR https://github.com/racker/reach/pull/12772

bradgignac:
Sounds good. The only suggestion I would make is to debounce the scroll handler and adding support for window resizing. Also, you might be interested in a library like http://github.hubspot.com/tether.

PR #41, opened 2015-09-10
Action menu

wpwood:
this PR address https://github.com/rackerlabs/canon-react/issues/15

### DIFF
--- a/src/Dropdown.jsx
+++ b/src/Dropdown.jsx
@@ -1,0 +1,45 @@
+import React from 'react';
+
+class Dropdown extends React.Component {
+  render() {
+    let style;
+
+    style = {float: 'left'};
+    return (
+      <div className={this._classes()} style={style}>
+        <ul className='rs-dropdown-menu visible'>
+          {this.props.children}
+        </ul>
+      </div>
+    );
+  }
+
+  _classes() {
+    let classes, dropdownTypes;
+
+    dropdownTypes = {
+      'primary': 'rs-nav-item rs-dropdown rs-primary-dropdown',
+      'utility': 'rs-nav-item rs-dropdown rs-utility-dropdown',
+      'action': 'rs-dropdown'
+    };
+
+    classes = [];
+    classes.push(this.props.className);
+
+    if (this.props.type && dropdownTypes[this.props.type]) {
+      classes.push(dropdownTypes[this.props.type]);
+    }
+
+    return classes.join(' ');
+  }
+}
+
+Dropdown.propTypes = {
+  type: React.PropTypes.oneOf(['primary', 'utility', 'action'])
+};
+
+Dropdown.defaultProps = {
+  type: 'action'
+};
+
+export default Dropdown;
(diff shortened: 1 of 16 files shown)


wpwood:
@willcumbie @jdrush89 @adampickeral 

I resurrected an old branch that Josh was working on and brought it up to date with the latest master. Thoughts?

adampickeral:
we have es6 support now. change it to use classes instead of `React.createClass`

src/DropdownItem.jsx
```diff
@@ -0,0 +1,63 @@
+var React = require('react');
+
+var DropdownItem = React.createClass({
+  propTypes: {
+    enabled: React.PropTypes.bool,
+    onClick: React.PropTypes.func,
+    type: React.PropTypes.string
```
adampickeral: seems like there are a limited number of types, so this should be `React.PropTypes.oneOf(['link', 'category', 'text'])`

src/DropdownItem.jsx
```diff
@@ -0,0 +1,63 @@
+var React = require('react');
+
+var DropdownItem = React.createClass({
+  propTypes: {
+    enabled: React.PropTypes.bool,
+    onClick: React.PropTypes.func,
```
adampickeral: .isRequired

src/Dropdown.jsx
```diff
@@ -0,0 +1,48 @@
...
+      classes.push(dropdownTypes[this.props.type]);
+    }
+
+    return classes.join(' ');
+  },
+
+  render: function () {
+    var style;
+
+    style = {float: 'left'};
```
adampickeral: why is this needed?

src/DropdownItem.jsx
adampickeral: actually, i wonder if this should be optional, and only applied if the type is `link`

src/DropdownItem.jsx
```diff
@@ -0,0 +1,63 @@
...
+    itemClasses = ['rs-dropdown-item'];
+    if (this.props.className) {
+      itemClasses.push(this.props.className);
+    }
+    if (!this.props.enabled) {
+      itemClasses.push('disabled');
+    }
+
+    return (
+      <li {...this.props} className={itemClasses.join(' ')} onClick={this._handleClick}>
```
adampickeral: `onClick` should only be added for links, right?

src/DropdownItem.jsx
wpwood: It was optional, and not supplied on the demo, resulting in an 'undefined' error when clicked. That's why I gave it a default value. I think it should be optional, and if not supplied, just a no-op.

wpwood:
@adampickeral I looked at only passing onClick for links. The problem is that (a) we pass props as a spread attribute so that other HTML attributes can be used, and (b) onClick is valid for other DOM elements. We would have to filter it out of props explicitly, and possibly break it for other uses. Filtering is not straightforward, either. There's an experimental ES7 syntax that JSX supports, but it requires a JSX flag.

Anyway, I think that it's more trouble than it's worth.

wpwood:
@willcumbie @jdrush89 @adampickeral Any other comments?

PR #42, opened 2015-09-25
Remove <trigger> element rendered by TooltipTrigger

adampickeral:


src/TooltipTrigger.jsx
```diff
@@ -22,6 +29,118 @@ class TooltipTrigger extends React.Component {
...
+
+    this._hideTooltip();
+    document.body.removeChild(this._containerDiv);
+  }
+
+  componentDidUpdate() {
+    this._toggleTooltip();
+  }
+
+  _toggleTooltip() {
```
willcumbie: "toggle" doesn't feel like the right word here--it's not like calling this will hide a tooltip if it's already shown. `updateVisibility`?

src/TooltipTrigger.jsx
```diff
@@ -37,20 +156,15 @@ class TooltipTrigger extends React.Component {
...
-        <Tooltip placement={this.props.placement}
-          isOpen={this._shouldShowTooltip()}
-          target={this._getTarget.bind(this)}
-          onMouseLeave={this._mouseLeavingTooltip.bind(this)}
-          onMouseOver={this._mouseEnteringTooltip.bind(this)}>
-          {this.props.content}
-        </Tooltip>
-      </trigger>
-    );
+    // <Tooltip placement={this.props.placement}
```
willcumbie: :scissors:

src/TooltipTrigger.jsx
adampickeral: :+1:

src/TooltipTrigger.jsx
adampickeral: :+1:

willcumbie:
LGTM

src/TooltipTrigger.jsx
```diff
@@ -37,20 +156,15 @@ class TooltipTrigger extends React.Component {
       ref: 'trigger'
```
jdrush89: Would 'target' be a better ref name?  It's a little confusing that the trigger component has a trigger ref.

adampickeral: hmm, i could go either way really. on one hand, the element is the actual trigger of the tooltip. on the other hand, target matches the name of the tether property that needs this ref.

adampickeral: i guess i'm going with trigger since i already merged haha. but i'm fine with renaming this in a future change.

PR #46, opened 2015-10-05
Allow additional classes on the popover footer

willcumbie:
We need this in Reach because our tests expect the button group to have "save-cancel" on it, and more generally because all components should be able to have arbitrary extra classes where needed.


underbyte:
:+1:

adampickeral:
this looks good to me.

i need to write a guide on contributing for this repo, but can you modify the commit message to follow these guidelines? https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines

you can look at some of the recent commits in master for examples

PR #48, opened 2015-10-06
fix(popover): reposition popover after render

willcumbie:
Create the tether before the popover content is rendered, and manually trigger a tether reposition after the popover has been rendered. This fixes an issue where autofocused controls cause bad scrolling behavior; previously, the popover content would be initially rendered in the wrong place and then quickly repositioned, causing autofocus to scroll to the wrong place (typically, to the bottom of the page). This fixes that behavior and makes the popover component compatible with autofocus.


underbyte:
LGTM

jdrush89:
:star:

adampickeral:
:+1:

PR #49, opened 2015-10-07
feat(PopoverBody): allow additional classNames on body

willcumbie:
Allow specification of additional popover body classNames via props. Previously there was no way to customize CSS classes on the body.


jdrush89:
:ship:

adampickeral:
:+1:

PR #50, opened 2015-10-07
chore(release): v 0.6.0

willcumbie:
1. Allow PopoverBody to receive additional class names as a property.
2. Fixed an issue where autoFocus would cause popovers to scroll out of view


adampickeral:
:shipit:

PR #51, opened 2015-10-12
fix(propTypes): use accurate propTypes

willcumbie:
Remove instances where 'children' is specified unnecessarily and 'className' is specified as an array rather than a string.

### DIFF
--- a/src/ButtonGroup.jsx
+++ b/src/ButtonGroup.jsx
@@ -17,9 +17,4 @@ class ButtonGroup extends React.Component {
   }
 }
 
-ButtonGroup.propTypes = {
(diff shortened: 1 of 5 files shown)


underbyte:
looks a lot better thanks for the fix!

wpwood:
@willcumbie Is 'any' right here for children? There's a 'node' PropType that is described as "Anything that can be rendered: numbers, strings, elements or an array (or fragment) containing these types."

willcumbie:
@wpwood `node` sounds better, I'll try that.

adampickeral:
:+1:

PR #53, opened 2015-10-14
Make clicks on dropdown items close the dropdown

adampickeral:
With the current implementation of dropdowns there is no way to hide the dropdown itself after clicking on an item of the dropdown. The desired behavior is that the dropdown should be hidden after it is clicked so a popover/navigation/some action can occur without the dropdown still being visible.


willcumbie:
LGTM

PR #57, opened 2015-10-26
Added a demo-build script.

wpwood:
Added a demo-build script. Fixed a typo on the demo page. Fixed a typo in the server script.

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -2,4 +2,4 @@
 
 # script/bootstrap: Resolve all dependencies required to develop and test canon-react components
 
-sudo npm install
+npm install
--- a/demo/DemoButtonSection.jsx
+++ b/demo/DemoButtonSection.jsx
@@ -73,7 +73,7 @@ var DemoButtonSection = React.createClass({
                 <td>
                   <Button type='action'>Actions</Button>
                 </td>
-                <td>Plus</td>
+                <td>Action</td>
                 <td><pre><code>{"<Button type='action'>Actions</Button>"}</code></pre></td>
               </tr>
             </tbody>
--- a/scripts/server
+++ b/scripts/server
@@ -2,6 +2,6 @@
 
 # script/server: Run a development server and watch for file changes while building demo components
 
-script/update
+scripts/update
 
-./node_modules/.bin/grunt watch & node node_modules/.bin/http-server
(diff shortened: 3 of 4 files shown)


wpwood:
@adampickeral @underbyte @willcumbie What say you?

scripts/setup
```diff
@@ -2,4 +2,4 @@
 
 # script/bootstrap: Resolve all dependencies required to develop and test canon-react components
 
-sudo npm install
```
underbyte: :+1: for not sudoing

underbyte:
LGTM

adampickeral:
scripts/server already builds the demo files and starts an http server serving them up

wpwood:
@adampickeral It didn't when I ran it this morning. No demo/bundle.js was created.

adampickeral:
@wpwood make a change and it will.

you can add this line `./node_modules/.bin/grunt demo-build` before this line `./node_modules/.bin/grunt watch & node node_modules/.bin/http-server` in `scripts/server` and it will build it when you first run the script

adampickeral:
i tried checking this out and i can't install without the sudo command

adampickeral:
i guess it works if i run `sudo scripts/setup`... but sudo is still required

wpwood:
@adampickeral Dunno. You shouldn't have to sudo npm unless you're installing something globally. Usually the issue is that you ran npm with sudo sometime in the past and forced a node module to be owned by root.

adampickeral:
ok i think i fixed my sudo issue. get rid of the demo-build script and update the server script and :+1:

wpwood:
@adampickeral I did

```
./node_modules/.bin/grunt demo-build && ./node_modules/.bin/grunt watch & node node_modules/.bin/http-server
```

so that the server command will exit if the demo-build fails.

wpwood:
I'd also love to rename the branch, but I think it will require a new pull request and losing the discussion :)

adampickeral:
cool, no need to rename the branch. do squash those two commits into one, and follow the commit message guidelines here: https://github.com/rackerlabs/canon-react/blob/master/CONTRIBUTING.md (this would be a 'chore' type of change). you can look at some of the recent merges for example commit messages

adampickeral:
LGTM

PR #58, opened 2015-11-03
chore(upgrade react): upgrade to react 0.14.2

adampickeral:
upgrade to using react 0.14.2

BREAKING CHANGE: this change requires users of the library to use react v 0.14.2

### DIFF
--- a/demo/DemoPopoverSection.jsx
+++ b/demo/DemoPopoverSection.jsx
@@ -1,4 +1,5 @@
 var React = require('react');
+var ReactDOM = require('react-dom');
 
 var Button = require('./Button');
 var DemoPopover = require('./DemoPopover');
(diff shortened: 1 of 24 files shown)


bradgignac:
Any reason you can't support old versions side-by-side? Support for the last two minor releases would be nice, in this case 0.13 and 0.14.

adampickeral:
@bradgignac we will. i'm not sure if there's a better method of doing it, but essentially any new features will need to be ported into a version based off of the current release (0.6.1)

adampickeral:
closing this for now. we'll support the two most recent versions of react (0.13, 0.14). since 0.13 api's are currently only deprecated, we don't need to make any changes until 0.15 is released.

Issue #55, opened 2015-10-15
Canon Facets


Jigarsolanki:

jdrush89:
PR https://github.com/jdrush89/canon-react/pull/1

This PR adds 4 components: FacetSection, Facet, Criteria, and FacetToggler.
FacetSection:
Renders the facet header and individual facet components.
Relies on a facetData object to generate facet components.  The object should be in the form:
```
{facetLabel: 
  {criteriaLabel: {count: number, iconClass: string, filter: func, disabled: opt_bool, hidden: opt_bool}}
}
```
For example:
```
facetData = {
      'TAGS': {
        'Engineering': {count: 8, filter: () => {}, disabled: true},
        'Sales': {count: 2, filter: () => {}},
        'Finance': {count: 0, filter: () => {}, hidden: true}
      },
      'STATUS': {
        'Error': {count: 2, filter: () => {}, iconClass: 'rs-status-error'},
        'Rebooting': {count: 3, filter: () => {}, iconClass: 'rs-status-warning'},
        'Active': {count: 5, filter: () => {}, iconClass: 'rs-status-ok'}
      }
    };
```
The FacetSection maintains the selected facets as state in an object of the same form.  When the selection changes, it fires a callback to its parent with the crtieria/facet labels and the filter function.

Facet:
This component renders a header and a list of criteria.  It expects a criteria object and selected criteria object, both in the form
```
{criteriaLabel: {count: number, iconClass: string, filter: func, disabled: opt_bool, hidden: opt_bool}
```
It will create criteria components from these objects, passing in the object values as props.  Facets truncate the list of criteria to 5 (by default) and display a FacetToggler to expand/collapse the list.  They also render a clear link when a criteria is selected that will clear all selections.

Criteria:
Criteria render a label and optional count in the form 'label (count)'.  They can be hidden, disabled, or given an icon class through props.  When clicked, they trigger a selection changed callback that should be used to apply the filter function associated with that criteria.

FacetToggler:
A simple link with arrow displaying '> more' or 'less ^'.  Triggers a toggle callback when clicked that should be used by a Facet to expand/collapse a list of criteria.

jdrush89:
There are two significant changes from this design I brought up in the PR and wanted to make a note of here.  I think it may be best to move the selected state out of the FacetsSection.  It will remove state from the component which makes it simpler, and encourages a single source of truth for the selected state (instead of having the selected filters being maintained outside and the state inside and tying to keep them in sync).

[...]

PR #61, opened 2015-11-30
Facets

jdrush89:
Adding the Facets components detailed in https://github.com/rackerlabs/canon-react/issues/55


adampickeral:
:shipit:

PR #62, opened 2015-12-14
converted remaining jsx files to ES6 syntax

underbyte:


underbyte:
@adampickeral ready for review!

src/ProgressBar.jsx
```diff
@@ -23,7 +23,7 @@ const TypeClasses = {
 class ProgressBar extends React.Component {
 
   _getSizeClass() {
-    var sizeClass;
+    let sizeClass;
```
adampickeral: for future reference, these don't really need to be `let`s. we don't get anything by using let here since the variable is declared at the top of the function

adampickeral:
looks good, squash those commits

adampickeral:
remember the git commit message format next time ;)

PR #63, opened 2015-12-19
chore(babel): upgrade babel to 6

underbyte:


.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: node_js
 
 node_js:
-  - "0.10"
+  - "5.3.0"
```
underbyte: We should at least standardize on version 5.0 of node. I got the babel error below only on node 0.10.

```
Running "browserify:test" (browserify) task
>> Error: module "babel-runtime/core-js/object/get-prototype-of" not found from "/Users/canon-react/transpiled/Criteria.js"
Warning: Error running grunt-browserify. Use --force to continue.
```

adampickeral:
have you tested the minified file to make sure it works with reach yet?

.travis.yml
adampickeral: you probably got that because of the different npm version that comes with node v5.

adampickeral:
squash commits and fix commit message

src/index.js
```diff
@@ -16,7 +16,7 @@ import ProgressBar from './ProgressBar';
 import StatusIndicator from './StatusIndicator';
 import TooltipTrigger from './TooltipTrigger';
 
-export default {
+window.canonReact = {
```
adampickeral: what if instead of `export default { }` you just did `export { }`

adampickeral:
```
export { Button, ButtonGroup, Criteria... }
```

package.json
```diff
@@ -39,23 +39,25 @@
...
   },
   "peerDependencies": {
     "react": ">=0.13.3"
   },
   "devDependencies": {
-    "babel": "^5.5.8",
-    "babel-eslint": "^4.1.1",
-    "babel-runtime": "^5.5.8",
-    "babelify": "^6.1.2",
+    "babel": "^6.3.26",
```
adampickeral: use exact versions for all the dependencies you've added/updated please

PR #64, opened 2016-01-13
fix(eslint): upgrade eslint

adampickeral:
the versions of eslint (and other lint plugins) were breaking with npm v 3. also removed a duplicate entry in the eslint config file


underbyte:
:+1:

PR #66, opened 2016-01-15
chore: upgrade react version

eduaquiles:
also added some rules on the eslint that helped us with the upgrade


adampickeral:
as much as i'd love to, we need to support the two most recent versions of react. once 0.15 comes out we can merge this

Issue #68, opened 2016-01-19
Upgrade Karma to 0.13


underbyte:
Since, we're switching to webpack it would be a good idea to do the upgrade during/after the switch. Because, we should be using karma's preprocessor to transpile and no point in having switch that too.

underbyte:
done in PR #71

PR #71, opened 2016-01-23
enchancement #68

underbyte:
upgrade karma and all karma dep
#68


adampickeral:
exact dependencies please.

better commit message:
```
chore(test upgrade): upgrade karma and jasmine

resolves issue #68
```

bradgignac:
For libraries, it's nice to have flexible dependencies so you don't force every consumer to a single version.

adampickeral:
interesting. these are all devDependencies, so it shouldn't matter for this, right?

on that point, I do worry about breaking changes with dependency upgrades when we don't specify an exact version, but if it's best practice for libraries to be more flexible, then we can do that with the other dependencies.

j00k:
Ideally everything but *major* releases should be backwards compatible but I'm not saying everyone npm contributor follows that guideline :smile_cat:

rogeriochaves:
:ship: :it:

bradgignac:
@adampickeral @j00k Consumers of any library should _always_ use `npm shrinkwrap` to lock their entire dependency tree.

bradgignac:
@adampickeral Dev dependencies matter because I'm pretty sure `npm install` installs them by default.

bradgignac:
@adampickeral @j00k One other thing: npm3 supports flattening the dependency tree, so you don't end up with multiple versions of the same library in use. This is great for server-side, but it is _very important_ for client-side builds to make sure you don't unnecessarily bloat your builds. When flattening dependency trees, it's very important to allow flexible versioning because the likelihood of a conflict is much higher.

adampickeral:
i'm pretty sure `devDependencies` don't work that way (what would be the point?). they only get installed if you run `npm install` from the root of that library package.

i tested this out by requiring `canon-react` as a dependency, running `npm install`, and ensuring that none of its `devDependencies` were installed (not alongside `canon-react` in `node_modules` nor in `node_modules` within `canon-react`)

bradgignac:
@adampickeral :ok_hand:

adampickeral:
:+1:

Issue #78, opened 2016-03-08
FacetsSection mutating props


jdrush89:
The FacetsSection mutates the object passed in for the selectedCriteria prop.  It should instead make a deep clone of this prop (must be deep as it's a nested object) and mutate that clone.

PR #79, opened 2016-03-08
fix(FacesSection): Clone selectedCriteria prop

jdrush89:
Perform a deep clone of the selectedCriteria prop, so the object referenced by the prop is not mutated

Fixes issue #78


adampickeral:
can you add a unit test that asserts that the passed in prop isn't changed?

jdrush89:
@adampickeral I added tests for each selection case where it was being modified.

PR #82, opened 2016-03-10
fix(Button): Fix HTML type of Button

### LINKED ISSUE #81: Button's react 'type' property conflicts with the html 'type' property
There are only three permissible types for an html button - 'submit', 'reset' and 'button'. Any other type is treated as a 'submit' type. This leads to unexpected behavior when using a non-submit button inside a form (Hitting enter triggers the onClick handler of the non-submit button). The correct behavior would be to set the type to 'button' in this case.

Could the react 'type' property be renamed to something else so as to not conflict with the html property of the same name?

ashimaathri:
There are only three permissible types for an html button - 'submit',
'reset' and 'button'. Any other type is treated as a 'submit' type. This
leads to unexpected behavior when using a non-submit button inside a
form (Hitting enter triggers the onClick handler of the non-submit
button). The correct behavior would be to set the type to 'button' in
this case and to 'submit' only when the button is intended as the
primary button of a form.

This fixes #81

### DIFF
--- a/src/Button.jsx
+++ b/src/Button.jsx
@@ -15,7 +15,15 @@ const BUTTON_TYPES = {
 
 class Button extends React.Component {
   render() {
-    let classes;
+    let classes, propsWithHTMLType;
+
+    propsWithHTMLType = Object.assign({}, this.props);
+    delete propsWithHTMLType.type;
+    if(['primary', 'login'].indexOf(this.props.type) !== -1) {
+      propsWithHTMLType.type = 'submit';
+    } else {
+      propsWithHTMLType.type = 'button';
+    }
 
     classes = classNames(
       this.props.className,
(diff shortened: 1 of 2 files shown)


src/Button.jsx
```diff
@@ -15,7 +15,15 @@ const BUTTON_TYPES = {
 
 class Button extends React.Component {
   render() {
-    let classes;
+    let classes, propsWithHTMLType;
+
+    propsWithHTMLType = Object.assign({}, this.props);
+    delete propsWithHTMLType.type;
```
akrikos: There's no need to delete this since you're reassigning it below.

akrikos:
This is an interesting 'hack' to make the primary be the button that always submits on 'enter' in a form field. The more I think about it, the more I like it. Note that if you have multiple primary buttons, the one that occurs first in the markup will be the one that submits.

I'm curious what @adampickeral thinks.

bradgignac:
I'm a little weary of this tying presentation elements of the button ("default", "primary", "login") to their behavior. Would it be better to pass through additional props to their native DOM equivalent and avoid custom properties that overlap? This is a pretty common pattern in most React components that wrap built-in elements.

adampickeral:
i agree with @bradgignac and @willcumbie. the button was already passing any extra props (including native DOM equivalents) along to the underlying button. like will said, the problem is that we have a collision with the native type property, and like brad said, we shouldn't tie the presentation of the button to its behavior. sooo....

change the `type` prop to `canonStyle`

this of course, will be a breaking change

ashimaathri:
Lol, I had the same question in #81. Maybe I should've asked for comments *before* submitting a PR :stuck_out_tongue: 
I'll make a new PR for the breaking change.

PR #83, opened 2016-03-11
fix(Button): Fix HTML type of Button

### LINKED ISSUE #81: Button's react 'type' property conflicts with the html 'type' property

ashimaathri:
There are only three permissible types for an html button - 'submit',
'reset' and 'button'. Any other type is treated as a 'submit' type. This
leads to unexpected behavior when using a non-submit button inside a
form (Hitting enter triggers the onClick handler of the non-submit
button). The correct behavior would be to set the type to 'button' in
this case and to 'submit' only when the button is intended as the
primary button of a form.
The button's type can now be set independently of the button's
canonStyle.

This fixes #81

### DIFF
--- a/demo/DemoButtonGroupSection.jsx
+++ b/demo/DemoButtonGroupSection.jsx
@@ -13,18 +13,18 @@ class DemoButtonGroupSection extends React.Component {
         </div>
         <div className='rs-detail-section-body'>
           <ButtonGroup>
-            <Button type='primary' enabled={true}>Primary</Button>
-            <Button type='secondary' enabled={true}>Secondary</Button>
-            <Button type='link' enabled={true}>Cancel</Button>
+            <Button canonStyle='primary' enabled={true}>Primary</Button>
+            <Button canonStyle='secondary' enabled={true}>Secondary</Button>
+            <Button canonStyle='link' enabled={true}>Cancel</Button>
             <ProcessingIndicator/>
           </ButtonGroup>
         </div>
         <div className='rs-detail-section-body'>
           <h3>Submitting State</h3>
           <ButtonGroup>
-            <Button type='primary' enabled={false}>Primary</Button>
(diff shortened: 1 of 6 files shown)


hannahstwinter:
You don't approve of free-style HTML? :smile_cat: The only Q I have is if ```canonStyle``` is the name we want... It's a Canon component so that seems a bit redundant. I'm unsure of better options right off of the top of my head... How would just ```style``` sit with you?

Jigarsolanki:
I liked this suggestion https://github.com/rackerlabs/canon-react/issues/81#issuecomment-194401033 better than what you have in this PR

adampickeral:
@hannahstwinter can't use `style` because it's an html attribute

underbyte:
`buttonType`? We can't use `style` because that will conflict with Element `style`.

adampickeral:
i'm fine with `canonStyle`

adampickeral:
bootstrap does something similar with its buttons: they use `bsStyle` (where `bs` stands for bootstrap) to specify the framework styling of the button. so i suggest `canonStyle`, `cStyle`, `cnStyle`...

adampickeral:
@Jigarsolanki no one liked what was done for #81, that's why she did it like this. the style of the button shouldn't imply its behavior

hannahstwinter:
@adampickeral haha I meant to search that before I commented... Should have done so.

hannahstwinter:
If we aren't abbreviating Canon anywhere else I say we just stick with canonStyle...

adampickeral:
:+1:

PR #87, opened 2016-04-04
bumped react version to 0.14.7

atarola:


rogeriochaves:
We had started a previous work on this too, although it should be outdated: https://github.com/rackerlabs/canon-react/pull/66

PR #88, opened 2016-04-14
chore: upgrade react version

rogeriochaves:
This merges the PRs #87 and #66

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -3,7 +3,8 @@
   <head>
     <link rel="stylesheet" type="text/css" href="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.css">
     <script type="application/javascript" src="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.js"></script>
-    <script src="https://fb.me/react-with-addons-0.13.3.min.js"></script>
+    <script src="https://fb.me/react-0.14.8.min.js"></script>
+    <script src="https://fb.me/react-dom-0.14.8.min.js"></script>
   </head>
   <body>
     <div class="rs-main">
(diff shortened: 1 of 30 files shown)


demo/index.html
```diff
@@ -3,7 +3,8 @@
   <head>
     <link rel="stylesheet" type="text/css" href="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.css">
     <script type="application/javascript" src="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.js"></script>
-    <script src="https://fb.me/react-with-addons-0.13.3.min.js"></script>
+    <script src="https://fb.me/react-0.14.8.min.js"></script>
```
wpwood: Not React 15? Is this because canon-react has to support previous React versions?

test/ButtonSpec.jsx
```diff
@@ -1,5 +1,6 @@
 import Button from '../transpiled/Button';
-import React from 'react/addons';
+import React from 'react';
+import ReactDOM from 'react-dom';
 let TestUtils = React.addons.TestUtils;
```
wpwood: Should this be 

```
import TestUtils from 'react-addons-test-utils';
```

like above?

demo/index.html
adampickeral: discussed IRL: we'll make one release that supports 0.14.8 then another one (with major version change) that supports 15

test/ButtonSpec.jsx
rogeriochaves: yes, this spec is the only one that does not work when I use the new test utils, still trying to figure out why while you review the rest of the pr :)

wpwood: Gotcha :)

wpwood:
Looks reasonable to me

:facepunch: :boom:

eduaquiles:
:+1:

PR #90, opened 2016-04-14
chore: upgrade react version

rogeriochaves:
now to v15.0.1, this was simpler since most work was done when upgrading to 0.14

the final bundle code didn't change one bit, only fixes to specs were made to cope with v15, so everybody working with 0.14 should also be able to work with this one


adampickeral:
:+1:

j00k:
:clap:

OneTechieMom:
Any chance we can get this merged?

adampickeral:
@OneTechieMom you shouldn't need this to be merged for canon-react to work with react v15

OneTechieMom:
I ended up upgrading to 15.0.1 since it doesn't have the unknown props issue but has the feature I need. Would like to get canon-react on the same version of react.

adampickeral:
we support the two most recent major versions of react, and right now we're able to do that without a branch in canon-react because react 0.14 compatible code is compatible with react 15 compatible code. So until react v16 comes out, we won't be bumping the required react version of canon-react (doing so would force all consumers of canon-react to upgrade to react v15 to get new features).

adampickeral:
upgraded in a separate PR

PR #91, opened 2016-04-21
fix(Facets): don't reference this.props in Facet constructor

jdrush89:
this.props is null in IE and that may be what is causing the Facets to break.  It could be that the props in general are null, either way there's no need to route though 'this' when we have the props as a param. 

edit:
tested on IE9, it works!


willcumbie:
:speedboat:

hannahstwinter:
L-🍑y-TM

PR #93, opened 2016-04-28
feat: optional tether prop for DropdownTrigger

hannahstwinter:
Introduces prop tether with value of 'right' or 'left' (default 'left') to DropdownTrigger.
Value of 'right' will tether the dropdown menu to the right of the trigger.

### DIFF
--- a/src/Dropdown.jsx
+++ b/src/Dropdown.jsx
@@ -9,9 +9,13 @@ const DROPDOWN_TYPES = {
 
 class Dropdown extends React.Component {
   render() {
-    let style, classes;
+    let dropdownStyle, menuStyle, classes;
 
-    style = { float: 'left' };
+    dropdownStyle = { float: 'left' };
+
+    if (this.props.alignment) {
+      menuStyle = { position: 'static' };
+    }
 
     classes = classNames(
       this.props.className,
@@ -19,8 +23,8 @@ class Dropdown extends React.Component {
     );
 
     return (
-      <div className={classes} style={style}>
-        <ul className='rs-dropdown-menu visible'>
+      <div className={classes} style={dropdownStyle}>
+        <ul className='rs-dropdown-menu visible' style={menuStyle}>
           { this._children() }
         </ul>
       </div>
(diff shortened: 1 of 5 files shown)


src/Dropdown.jsx
```diff
@@ -9,18 +9,22 @@ const DROPDOWN_TYPES = {
...
 class Dropdown extends React.Component {
   render() {
-    let style, classes;
+    let dropdownStyle, menuStyle, classes;
 
-    style = { float: 'left' };
+    dropdownStyle = { float: 'left' };
+
+    if (this.props.tether) {
+      menuStyle = { position: 'static' };
```
hannahstwinter: Absolute positioned elements with absolute positioned children do not change size in relation to their children. The _containerDiv created in the DropdownTrigger was not expanding to the size of the menu. In accord, the container div had a size of 0 and was positioned at the top left corner of the menu, so tethering on any corner of the container div would always translate to the top left corner of the menu.

willcumbie:
I'm not crazy about the property name "tether." What about "position" or "alignment"?

adampickeral:
i would rename the `tether` prop to `placement`.

hannahstwinter:
I could go for alignment - it would read better against the concept of Tether in the component, the Tether instance, etc.

adampickeral:
`placement` is what the popover uses

hannahstwinter:
I worry 'placement' of right might convey that the menu should be placed to the right of the button (implying the tether point is on the left).

adampickeral:
`alignment` works for me, makes sense in this case. the dropdown is aligned to the right or left of the trigger

src/Dropdown.jsx
jdrush89: does it hurt to always have this style?

hannahstwinter: @jdrush89 I don't know... Also, I'm not 100% certain how to test that thoroughly enough to feel ok with making the change universal. 😿

adampickeral:
LGTM!

src/Dropdown.jsx
jdrush89: I think we're only using this in a few places in reach, I can help test when you bump the version if you want.  Without it we're relying on the 0 width container defaulting to the top left and I don't know if that behavior could change.

wpwood:
LGTM2

🌸 🍶

Issue #77, opened 2016-02-25
Exposing Tethering low level parameters in Popovers


MACSkeptic:
Recently we implemented a popover in Janus that looks like this:

![image](https://cloud.githubusercontent.com/assets/53897/13319353/5f6d87b8-dbc1-11e5-94aa-026904b6a5b4.png)

We couldn't make that work with the exposed tethering abstractions from https://github.com/rackerlabs/canon-react/blob/master/src/Popover.jsx#L105.

The positioning on the screenshot is done by tricking the algorithm with a absolutely positioned ghost element.

I believe it would be possible to do that if we could set the tether config to:

```
     attachment: 'top left',
     targetAttachment: 'bottom left',
```

That said, I'd expect more scenarios could also benefit from this. I think the abstract tethering named configurations are good, but would be great if we could access/influence the raw settings for Popover positioning.

adampickeral:
resolved by #97

PR #97, opened 2016-05-02
feat(popover): Override tether config

adampickeral:
Allow users to manually override all tether config options. This allows users to have better control over placement of popovers for special cases.

Resolves issue #77

### DIFF
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -149,8 +149,8 @@ class Popover extends React.Component {
         };
     }
 
-    if (this.props.offset) {
-      tetherConfig.offset = this.props.offset;
+    if (this.props.tetherConfig) {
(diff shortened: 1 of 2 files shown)


willcumbie:
I wondered whether this would become an issue; I guess so. Seems like a good fix.

akrikos:
This looks like a breaking change. Do you know how many usages you'll have to fix?

adampickeral:
@akrikos no changes are needed in reach. I forget now why the `offset` was needed in the first place. I did forget to add that this is a breaking change in the commit message though. I'll update that

MACSkeptic:
👍  this is super helpful

MACSkeptic:
this is the madness that will be removed thanks to this change:

image

PR #98, opened 2016-05-03
Fix(Popover) don't add rs-popover class if it already exists

jdrush89:
Noticed the container was getting an extra 'rs-popover' class every time its props updated.  Adding a check for the class so it won't be added if it already exists.

### DIFF
--- a/src/Popover.jsx
+++ b/src/Popover.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Tether from 'tether';
 import PopoverBackground from './PopoverBackground';
+import classNames from 'classnames';
 
 class Popover extends React.Component {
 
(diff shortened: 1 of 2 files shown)


src/Popover.jsx
```diff
@@ -79,7 +80,9 @@ class Popover extends React.Component {
 
     this._backgroundDiv.style.display = 'block';
     ReactDOM.render(<PopoverBackground isModal={ this.props.isModal } onRequestClose={this.props.onRequestClose} />, this._backgroundDiv);
-    this._containerDiv.className += ' rs-popover';
+
+    let containerClasses = this._containerDiv.className.split(' ');
+    this._containerDiv.className = classNames(containerClasses, {'rs-popover': containerClasses.indexOf('rs-popover') < 0});
```
wpwood: Are you sure that you need to split and check this? I would be surprised if the classnames library didn't do the right thing if 'rs-popover' was already in the list of classes.

wpwood: Tested it out. I retract my statement, classNames will totally just add the class multiple times.

wpwood:
LGTM

👊 💥

adampickeral:
:+1:

src/Popover.jsx
hannahstwinter: #includes is too fancy for IE - I always forget that...

hannahstwinter:
L 🗽 TM

Issue #99, opened 2016-05-10
Context not available in popovers


adampickeral:
Context is not available in popovers, and breaks links within those popovers when using a react router. Rendering the popovovers with `unstable_renderSubtreeIntoContainer` will fix the issue, similar to how react-modal is using it: https://github.com/reactjs/react-modal/blob/master/lib/components/Modal.js#L7

adampickeral:
Fixed by #101

PR #101, opened 2016-05-10
fix(popover): popover missing context from parent

adampickeral:
render popover subtree into a container so it gets the context from its parent component

addresses issue #99


MACSkeptic:
👍  w00t

PR #103, opened 2016-06-01
Pass parent context to tooltip and dropdown

Jigarsolanki:


MACSkeptic:
👍  sorry I missed this, ended up implemented it for TooltipTrigger as part of https://github.com/rackerlabs/canon-react/pull/117

src/DropdownTrigger.jsx
```diff
@@ -88,7 +88,7 @@ class DropdownTrigger extends React.Component {
       }
     );
 
-    this._dropdownNode = ReactDOM.render(dropdown, this._containerDiv);
+    this._dropdownNode = ReactDOM.unstable_renderSubtreeIntoContainer(dropdown, this._containerDiv);
```
MACSkeptic: I think this takes an extra "this" parameter

Jigarsolanki: Yes! Thanks

src/DropdownTrigger.jsx
```diff
@@ -88,7 +88,7 @@ class DropdownTrigger extends React.Component {
       }
     );
 
-    this._dropdownNode = ReactDOM.render(dropdown, this._containerDiv);
+    this._dropdownNode = ReactDOM.unstable_renderSubtreeIntoContainer(this, dropdown, this._containerDiv);
```
wpwood: Is there any way to test this? Sure would be nice to have a test that it actually passed the context, since I'm assuming that's what would not have happened, if you had left the 'this' out.

Jigarsolanki: The test had failed when I left it out.

MACSkeptic: we test contexts like this in some places: https://github.com/racker/apollo/blob/master/app/containers/App.test.js#L64

wpwood:
LGTM

🍶 🍜 🍣

PR #105, opened 2016-06-07
feat: added form and field components

willcumbie:
Added Form, FormField, and related components to encapsulate Canon forms and form validation messages

Changes are based on the Apollo changes here, updated to Canon-React patterns: https://github.com/racker/apollo/pull/45/files

I've removed FormButton and related, since they reproduce button logic we already have in other places.

### DIFF
--- a/demo/DemoForm.jsx
+++ b/demo/DemoForm.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+
+import Form from './Form';
+import FormField from './FormField';
+import DetailsSection from './DetailsSection';
+
+class DemoForm extends React.Component {
+  render() {
+    return (
+      <DetailsSection title="Form">
+        <Form size="large">
+          <FormField  label="Standard Input" help="This is a field with no particular status">
+            <input type="text" />
+          </FormField>
+          <FormField error="This input is not valid" label="Error Field">
+            <input type="text" value="invalid text" />
+          </FormField>
+          <FormField success="This input is valid" label={ <span><i>Fixed</i> Error Field</span>}>
+            <select>
+               <option>Option One</option>
+               <option>Option Two</option>
+             </select>
+          </FormField>
+        </Form>
+      </DetailsSection>
+    );
+  }
+}
+
+export default DemoForm;
--- a/demo/DemoFormPopover.jsx
+++ b/demo/DemoFormPopover.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+
+import FormField from './FormField';
+import FormPopover from './FormPopover';
+
+class DemoPopover extends React.Component {
+  render() {
+    return (
+      <FormPopover {...this.props} >
+        <FormField label="Field">
+          <input type="text" />
+        </FormField>
+      </FormPopover>
+    );
+  }
+}
+
+export default DemoPopover;
--- a/src/Form.jsx
+++ b/src/Form.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const SIZE_CLASSES = {
+  'xsmall': 'rs-form-xsmall',
+  'small': 'rs-form-small',
+  'medium': 'rs-form-medium',
+  'large': 'rs-form-large',
+  'xlarge': 'rs-form-xlarge'
+};
+
+export default class Form extends React.Component {
+  render() {
+    let classes, sizeClass;
+
+    sizeClass = SIZE_CLASSES[this.props.size];
+    classes = classNames(
+      'rs-form-horizontal',
+      sizeClass,
+      this.props.create ? 'rs-form-create' : undefined,
+      this.props.className
+    );
+
+    return (
+      <form {...this.props} className={classes}>
+        {this.props.children}
+      </form>
+    );
+  }
+}
+
+Form.propTypes = {
+  create: React.PropTypes.boolean,
+  className: React.PropTypes.string,
+  size: React.PropTypes.oneOf(Object.keys(SIZE_CLASSES)),
+  children: React.PropTypes.node.isRequired
+};
--- a/src/FormField.jsx
+++ b/src/FormField.jsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import FormFieldHelp from './FormFieldHelp';
+import FormFieldValidationBlock from './FormFieldValidationBlock';
+import classnames from 'classnames';
+
+export default class FormField extends React.Component {
+  render() {
+    return (
+      <div className={classnames('rs-control-group', { 'error': this.props.error, 'success': this.props.success })}>
+        <label className="rs-control-label">{this.props.label}</label>
+        <div className="rs-controls">
+          {this.props.children}
+          <FormFieldHelp help={this.props.help} />
+          <FormFieldValidationBlock value={this.props.error || this.props.success} />
+        </div>
+      </div>
+    );
+  }
+}
+
+FormField.propTypes = {
+  error: React.PropTypes.string,
+  success: React.PropTypes.string,
+  help: React.PropTypes.node,
+  label: React.PropTypes.node.isRequired,
+  children: React.PropTypes.node.isRequired
+};
--- a/src/FormFieldHelp.jsx
+++ b/src/FormFieldHelp.jsx
@@ -1,0 +1,13 @@
+import React from 'react';
+
+export default class FormFieldHelp extends React.Component {
+  render() {
+    return this.props.help ? (
+      <span className="rs-help-block">{this.props.help}</span>
+    ) : null;
+  }
+}
+
+FormFieldHelp.propTypes = {
+  help: React.PropTypes.node
+};
--- a/src/FormPopover.jsx
+++ b/src/FormPopover.jsx
@@ -1,0 +1,88 @@
+import React from 'react';
+import Button from './Button';
+import Form from './Form';
+import FormFieldValidationBlock from './FormFieldValidationBlock';
+import Popover from './Popover';
+import PopoverOverlay from './PopoverOverlay';
+import PopoverBody from './PopoverBody';
+import PopoverFooter from './PopoverFooter';
+import ProcessingIndicator from './ProcessingIndicator';
+
+class FormPopover extends React.Component {
+  render() {
+    let submitComponent, cancelComponent, popoverProps;
+
+    submitComponent = React.cloneElement(
+      this.props.submitButton,
+      { onClick: this._submit.bind(this), enabled: !this.props.processing }
+    );
+    cancelComponent = React.cloneElement(
+      this.props.cancelButton,
+      { onClick: this._cancel.bind(this), hidden: this.props.processing }
+    );
+
+    popoverProps = Object.assign({}, this.props);
+    delete popoverProps.children;
+
+    return (
+      <Popover {...this.props} >
+        <PopoverOverlay>
+          <Form onsubmit={this.props.onSubmit} size={this.props.formSize} >
+            <PopoverBody>
+              {this.props.children}
+            </PopoverBody>
+            <PopoverFooter>
+              {submitComponent}
+              {cancelComponent}
+              <ProcessingIndicator hidden={ !this.props.processing }/>
+              <FormFieldValidationBlock value={this.props.error} />
+            </PopoverFooter>
+          </Form>
+        </PopoverOverlay>
+      </Popover>
+    );
+  }
+
+  _cancel(event) {
+    if (event && event.preventDefault) {
+      event.preventDefault();
+    }
+    this.props.onRequestClose();
+  }
+
+  _submit(event) {
+    if (event && event.preventDefault) {
+      event.preventDefault();
+    }
+    this.props.onSubmit(event);
+  }
+}
+
+FormPopover.propTypes = {
+  // form
+  formSize: React.PropTypes.string,
+  error: React.PropTypes.node,
+  processing: React.PropTypes.bool.isRequired,
+  onSubmit: React.PropTypes.func.isRequired,
+  // popover layout
+  cancelButton: React.PropTypes.element.isRequired,
+  submitButton: React.PropTypes.element.isRequired,
+  // popover content
+  children: React.PropTypes.node.isRequired,
+  // popover
+  isOpen: React.PropTypes.bool.isRequired,
+  target: React.PropTypes.string.isRequired,
+  placement: React.PropTypes.string.isRequired,
+  onRequestClose: React.PropTypes.func.isRequired
+};
+
+FormPopover.defaultProps = {
+  error: null,
+  cancelButton: <Button canonStyle="link">Cancel</Button>,
+  submitButton: <Button canonStyle="primary">Submit</Button>,
+  processing: false,
+  isOpen: true,
+  placement: 'right'
+};
+
+export default FormPopover;
--- a/test/FormFieldHelpSpec.jsx
+++ b/test/FormFieldHelpSpec.jsx
@@ -1,0 +1,24 @@
+import FormFieldHelp from '../transpiled/FormFieldHelp';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('FormFieldHelp', () => {
+  it('displays the help message', () => {
+    let message, formFieldHelp;
+
+    formFieldHelp = TestUtils.renderIntoDocument(<FormFieldHelp help="test message" />);
+    message = TestUtils.findRenderedDOMComponentWithClass(formFieldHelp, 'rs-help-block');
+    expect(message.getDOMNode().textContent).toBe('test message');
+
+    ReactDOM.unmountComponentAtNode(ReactDOM.findDOMNode(formFieldHelp).parentNode);
+  });
+
+  it('returns null if no field is passed to it', () => {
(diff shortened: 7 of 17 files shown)


src/Form.jsx
```diff
@@ -0,0 +1,37 @@
...
+  'xlarge': 'rs-form-xlarge'
+};
+
+export default class Form extends React.Component {
+  render() {
+    let classes, sizeClass;
+
+    sizeClass = SIZE_CLASSES[this.props.size];
+    classes = classNames(
+      'rs-form-horizontal',
```
willcumbie: I wasn't sure about this class. Canon doesn't seem to support simply `rs-form`, and from what I can tell there is no `rs-form-vertical`. If there are non-horizontal forms in the future, this will have to change--but most likely a change that big would necessitate breaking changes to this component whether or not we add `rs-form-horizontal` by default.

src/Form.jsx
```diff
@@ -0,0 +1,37 @@
...
+        {this.props.children}
+      </form>
+    );
+  }
+}
+
+Form.propTypes = {
+  create: React.PropTypes.boolean,
+  className: React.PropTypes.string,
+  size: React.PropTypes.string,
```
hannahstwinter: Should you do a oneOf() here?

src/FormField.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import FormFieldHelp from './FormFieldHelp';
+import FormFieldValidationBlock from './FormFieldValidationBlock';
+
+export default class FormField extends React.Component {
+  render() {
+    return (
+      <div className={`rs-control-group ${this.props.error && 'error'} ${this.props.success && 'success'}`}>
```
hannahstwinter: I think I would prefer classNames here. I would at least prefer consistency, whichever one we're leaning more towards in Canon-React.

src/FormField.jsx
```diff
@@ -0,0 +1,26 @@
...
+          <FormFieldValidationBlock value={this.props.error || this.props.success} />
+        </div>
+      </div>
+    );
+  }
+}
+
+FormField.propTypes = {
+  error: React.PropTypes.string,
+  success: React.PropTypes.string,
```
hannahstwinter: How would you feel about a custom validator to ensure that only one of 'error' or 'success' props get passed in here?

demo/DemoForm.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+import Form from './Form';
+import FormField from './FormField';
+
+class DemoForm extends React.Component {
+  render() {
+    return (
+      <div className='rs-detail-section'>
```
hannahstwinter: Why are you not using the FormSection component here?

test/FormFieldHelpSpec.jsx
```diff
@@ -0,0 +1,29 @@
+import FormFieldHelp from '../transpiled/FormFieldHelp';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('FormFieldHelp', () => {
+  let formFieldHelp;
+
+  beforeEach(() => {
+    formFieldHelp = TestUtils.renderIntoDocument(<FormFieldHelp />);
```
hannahstwinter: I would either move this into the assertion on 23 or extract the two cases into describe blocks and do the setup in a before each. I like the latter more, I think.

test/FormFieldSpec.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+describe('FormField', () => {
+  let formField;
+
+  beforeEach(() => {
+    formField = TestUtils.renderIntoDocument(<FormField>test</FormField>);
+  });
+
+  afterEach(() => {
+    ReactDOM.unmountComponentAtNode(ReactDOM.findDOMNode(formField).parentNode);
```
hannahstwinter: You're not unmounting consistently across other specs. Do we not automate that cleanup in Canon-React?

test/FormFieldSpec.jsx
```diff
@@ -0,0 +1,76 @@
+import FormField from '../transpiled/FormField';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('FormField', () => {
+  let formField;
+
+  beforeEach(() => {
+    formField = TestUtils.renderIntoDocument(<FormField>test</FormField>);
```
hannahstwinter: You're not touching this instance of formField - you're re-rendering it in each spec or beforeEach function within your describe blocks.

test/FormFieldSpec.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+    it('adds the message to the validation block', () => {
+      let message;
+
+      message = TestUtils.findRenderedDOMComponentWithClass(formField, 'rs-validation-block');
+      expect(message.getDOMNode().textContent).toBe(' Test error message');
+    });
+  });
+
+  describe('when there is a fixed error', () => {
```
hannahstwinter: Really this is just 'when there is success', without inferring how this will be consumed. Wow, it's early for nit picking, sorry. ;)

test/FormFieldValidationBlockSpec.jsx
```diff
@@ -0,0 +1,29 @@
+import FormFieldValidationBlock from '../transpiled/FormFieldValidationBlock';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('FormFieldValidationBlock', () => {
+  let formFieldValidationBlock;
+
+  beforeEach(() => {
+    formFieldValidationBlock = TestUtils.renderIntoDocument(<FormFieldValidationBlock />);
```
hannahstwinter: The same as in FormFieldHelp applies here, IMO.

test/FormSectionSpec.jsx
```diff
@@ -0,0 +1,34 @@
+import FormSection from '../transpiled/FormSection';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('FormSection', () => {
+  let formSection;
+
+  beforeEach(() => {
+    formSection = TestUtils.renderIntoDocument(<FormSection />);
```
hannahstwinter: You're not touching this instance of formSection.

test/FormSpec.jsx
```diff
@@ -0,0 +1,42 @@
+import Form from '../transpiled/Form';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-addons-test-utils';
+
+describe('Form', () => {
+  let form;
+
+  beforeEach(() => {
+    form = TestUtils.renderIntoDocument(<Form>test</Form>);
```
hannahstwinter: You're not touching this instance of form.

src/FormField.jsx
willcumbie: I doubt it would be worth the trouble, to be honest. It seems intuitively obvious that the component shouldn't receive both; if it does, it'll render inconsistently/strangely, cluing the user in that something is wrong.

hannahstwinter: Seems reasonable. I bring it up b/c I try to abide by the practice of assuming that users are going to try their hardest to break something, whether or not they are aware of it.

src/Form.jsx
adampickeral: yes, and probably a default value for it? not sure if the size is required or not

willcumbie: Size is not required. oneOf sounds good.

src/FormFieldHelp.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export default class FormFieldHelp extends React.Component {
```
adampickeral: could this be a function instead of a class? same question goes for all of these really.

I'm also open to creating a separate issue to convert all the existing classes to functions (that can be converted) if you don't want to change these in this PR

src/FormSection.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+export default class FormSection extends React.Component {
```
adampickeral: What makes this a form section and not just a details section?

src/FormFieldHelp.jsx
willcumbie: Yeah, I think I'll punt on this and do a bunch of component-to-function conversions in a follow-up request.

demo/DemoFormPopover.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import FormField from './FormField';
+import FormPopover from './FormPopover';
+
+class DemoPopover extends React.Component {
+
```
hannahstwinter: 🙈

willcumbie: humphrey bogart - facepalm

src/FormField.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import FormFieldHelp from './FormFieldHelp';
+import FormFieldValidationBlock from './FormFieldValidationBlock';
+import classnames from 'classnames';
+
+export default class FormField extends React.Component {
+  render() {
+    return (
+      <div className={classnames('rs-control-group', this.props.error && 'error', this.props.success && 'success')}>
```
hannahstwinter: classnames has specific syntax for conditionally adding a class - ```classNames('foo', { bar: true, baz: false})``` returning ```'foo bar'```

src/FormPopover.jsx
```diff
@@ -0,0 +1,91 @@
...
+import Button from './Button';
+import Form from './Form';
+import FormFieldValidationBlock from './FormFieldValidationBlock';
+import Popover from './Popover';
+import PopoverOverlay from './PopoverOverlay';
+import PopoverBody from './PopoverBody';
+import PopoverFooter from './PopoverFooter';
+import ProcessingIndicator from './ProcessingIndicator';
+
+class FormPopover extends React.Component {
```
hannahstwinter: I'm not certain about this component entirely... Maybe it's fine but I'm not sure that we will be able to use it without exposing a number of other optional/customizable features (we use a Reach specific ControlsError on each submittable popover, for instance). Additionally, we could use some sort of analytics in our submittable popover component. 
I'm working on one within Reach right now that addresses those things in addition to tying directly to our controllers. I'll ping you when it's close.

src/FormPopover.jsx
```diff
@@ -0,0 +1,91 @@
...
+      this.props.submitButton,
+      { onClick: this._submit.bind(this), enabled: !this.props.processing }
+    );
+    cancelComponent = React.cloneElement(
+      this.props.cancelButton,
+      { onClick: this._cancel.bind(this), hidden: this.props.processing }
+    );
+
+    popoverProps = Object.assign({}, this.props);
+    delete popoverProps.children;
```
hannahstwinter: What's this about?

willcumbie: We want to allow the user to pass arbitrary props to the Popover component, so we pass all of props, but we can't also pass the `children` prop, since we're already rendering it elsewhere (it'd get rendered twice).

src/FormPopover.jsx
willcumbie: You can replace the buttons, which seem like the big thing that would differ. What particular other things do you think we'll need to customize? I can add those in as well.

hannahstwinter: I'm not sure... This could very well be completely sufficient - I feel like it addresses most of the normal cases we bump into. I would have to play with it more to be sure, and try it in places like our multiple popover patterns (create security group then add rule to group being one), though again, I think it might cover all of the bases as it stands. :+1:

hannahstwinter:
L🐹TM

lucasrose:
cloned repo to check out your demo as well, LGTM!

demo/DemoForm.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+import Form from './Form';
+import FormField from './FormField';
+import DetailsSection from './DetailsSection';
+
+class DemoForm extends React.Component {
+  render() {
+    return (
+      <DetailsSection title="Form">
+        <Form size="large">
```
MACSkeptic: I like the size abstraction

src/FormField.jsx
MACSkeptic: I think validating is actually a good idea... probably not too farfetched to assume that `success` and `error` might end up co-existing if there are issues with stale state/props somewhere.

src/FormFieldHelp.jsx
MACSkeptic: 👍

src/Form.jsx
```diff
@@ -0,0 +1,37 @@
...
+    sizeClass = SIZE_CLASSES[this.props.size];
+    classes = classNames(
+      'rs-form-horizontal',
+      sizeClass,
+      this.props.create ? 'rs-form-create' : undefined,
+      this.props.className
+    );
+
+    return (
+      <form className={classes}>
```
MACSkeptic: why aren't we handling submit here instead of using on-click on the button?

using the button instead of submit makes things like hitting enter for submit not work out of the box

src/FormPopover.jsx
```diff
@@ -0,0 +1,88 @@
...
+  // popover
+  isOpen: React.PropTypes.bool.isRequired,
+  target: React.PropTypes.string.isRequired,
+  placement: React.PropTypes.string.isRequired,
+  onRequestClose: React.PropTypes.func.isRequired
+};
+
+FormPopover.defaultProps = {
+  error: null,
+  cancelButton: <Button canonStyle="link">Cancel</Button>,
```
MACSkeptic: cancel is a button?

test/FormSpec.jsx
```diff
@@ -0,0 +1,38 @@
...
+    form = TestUtils.renderIntoDocument(<Form size="xsmall">test</Form>);
+    expect(ReactDOM.findDOMNode(form)).toHaveClass('rs-form-xsmall');
+  });
+
+  it('does not add invalid size classes', () => {
+    form = TestUtils.renderIntoDocument(<Form size="ginormous">test</Form>);
+    expect(ReactDOM.findDOMNode(form)).not.toHaveClass('rs-form-ginormous');
+  });
+
+  it('adds the create class', () => {
```
MACSkeptic: philosophical question here: when would you use a "create" form or a "non-create" form ?

willcumbie: Canon defines create forms as particularly wide forms that fill a page, rather than popovers. For create forms, labels are right-aligned. More information here: http://rackerlabs.github.io/canon/ui-components/#forms

src/FormPopover.jsx
willcumbie: Yes: http://rackerlabs.github.io/canon/ui-components/#buttons

test/FormSpec.jsx
MACSkeptic: that explains a lot, thanks

src/FormPopover.jsx
```diff
@@ -0,0 +1,88 @@
...
+      { onClick: this._cancel.bind(this), hidden: this.props.processing }
+    );
+
+    popoverProps = Object.assign({}, this.props);
+    delete popoverProps.children;
+
+    return (
+      <Popover {...this.props} >
+        <PopoverOverlay>
+          <Form onSubmit={this.props.onSubmit} size={this.props.formSize} >
```
MACSkeptic: would you foresee a form inside a form type of issue?

I almost hit that on apollo the other day because of inline actions inside a form popover. We ended up going a different path so I didn't solve the problem, but it might be something to consider since this approach forces you to wrap the entire content of the popover on a form.

if the answer for that is "not likely to be a problem", that would add to my other comment in favour of preferring submit vs button click handlers

willcumbie: I don't think a form inside a form is very likely. I'm planning to explore handling both click and submit, in case we get into situations with unusual button components. I don't anticipate it being difficult to listen for both.

src/Form.jsx
willcumbie: Figured out why this was working: the submit buttons in Canon are actual buttons with formatting, not `<input type="submit">` components.

On the off-chance that someone needs to use submit inputs, this ought to still work. I'll modify this class to make that possible.

src/FormField.jsx
willcumbie: Yeah, that's far. I'm going to keep that out of this PR, though. It'll either require pulling in an external dependency or adding a utility methods file (for the new validator method) to canon-react, neither of which I want to do here.

src/Form.jsx
hannahstwinter: Crud - I wish I had given this a second look. 'rs-form-horizontal' is the class we put on the form to align the label element and the field element next to one another. There is no complimenting class (such as the hypothetical 'rs-form-veritcal'), you simply leave it off and then the label and the form field stack on top of one another. This is a very common pattern in our single element popovers (rename being the most common single field popover, I think).
See the 'Top aligned field labels' section under http://rackerlabs.github.io/canon/ui-components/#popover

willcumbie: Yeah, the example there just doesn't use any rs-form class at all, so it wouldn't change this at all.

hannahstwinter: So you're thinking we would just skip this wrapper altogether and just wrap it in a classless form tag? That actually makes perfect sense now that I look at this again and realize that all this component really does is add classes to the form tag. :X

PR #108, opened 2016-06-17
feat(details-section):allow collapsibility

lucasrose:
Added property to DetailsSection component isCollapsible,
which when passed true makes the section collapsible.

### DIFF
--- a/src/DetailsSection.jsx
+++ b/src/DetailsSection.jsx
@@ -6,7 +6,7 @@ export default class DetailsSection extends React.Component {
     super(props);
 
     if (props.isCollapsible === true) {
-      this.state = { isCollapsed: false };
+      this.state = { isCollapsed: props.defaultCollapsed };
     }
   }
 
@@ -47,9 +47,11 @@ DetailsSection.propTypes = {
   title: React.PropTypes.node.isRequired,
   children: React.PropTypes.node,
   className: React.PropTypes.string,
-  isCollapsible: React.PropTypes.bool
+  isCollapsible: React.PropTypes.bool,
+  defaultCollapsed: React.PropTypes.bool
 };
 
 DetailsSection.defaultProps = {
-  isCollapsible: false
+  isCollapsible: false,
+  defaultCollapsed: false
 };
--- a/src/DetailsSection.jsx
+++ b/src/DetailsSection.jsx
@@ -1,11 +1,31 @@
 import React from 'react';
+import classnames from 'classnames';
 
 export default class DetailsSection extends React.Component {
+  constructor(props) {
+    super(props);
+
+    if (props.isCollapsible === true) {
+      this.state = { changedClass: 'expanded' };
+    } else {
+      this.state = { changedClass: undefined };
+    }
+  }
+
+  collapseSection() {
+    if (this.state.changedClass === 'expanded') {
+      this.setState({ changedClass: 'collapsed' });
+    } else if (this.state.changedClass === 'collapsed') {
+      this.setState({ changedClass: 'expanded' });
+    }
+  }
+
   render() {
     return (
-      <div className="rs-detail-section">
+      <div isCollapsible={ this.props.isCollapsible } onClick={ this.collapseSection.bind(this) } className={ classnames('rs-detail-section rs-collapsible-section', this.state.changedClass) }>
         <div className="rs-detail-section-header">
-          <div className="rs-detail-section-title">{this.props.title}</div>
+          { this.props.isCollapsible ? <div className="rs-caret"></div> : null}
+          <div className="rs-detail-section-title">{ this.props.title }</div>
         </div>
         <div className="rs-detail-section-body">
           {this.props.children}
@@ -17,5 +37,11 @@ export default class DetailsSection extends React.Component {
 
 DetailsSection.propTypes = {
   title: React.PropTypes.node.isRequired,
-  children: React.PropTypes.node
+  children: React.PropTypes.node,
+  className: React.PropTypes.string,
+  isCollapsible: React.PropTypes.bool
+};
+
+DetailsSection.defaultProps = {
+  isCollapsible: false
 };
--- a/src/DetailsSection.jsx
+++ b/src/DetailsSection.jsx
@@ -6,28 +6,36 @@ export default class DetailsSection extends React.Component {
     super(props);
 
     if (props.isCollapsible === true) {
-      this.state = { changedClass: 'expanded' };
-    } else {
-      this.state = { changedClass: undefined };
+      this.state = { isCollapsed: false };
     }
   }
 
-  collapseSection() {
-    if (this.state.changedClass === 'expanded') {
-      this.setState({ changedClass: 'collapsed' });
-    } else if (this.state.changedClass === 'collapsed') {
-      this.setState({ changedClass: 'expanded' });
+  toggleCollapse() {
+    if (this.props.isCollapsible) {
+      this.setState({isCollapsed: !this.state.isCollapsed});
     }
   }
 
   render() {
+    var collapsedClassName, showCaret;
+
+    showCaret = false;
+    if (this.props.isCollapsible) {
+      showCaret = true;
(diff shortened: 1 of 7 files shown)


demo/DemoDetailSection.jsx
```diff
@@ -6,40 +6,36 @@ import DetailList from './DetailList';
 import DetailItem from './DetailItem';
 import DetailItemKey from './DetailItemKey';
 import DetailItemValue from './DetailItemValue';
+import DetailsSection from './DetailsSection';
 
 class DemoDetailSection extends React.Component {
   render() {
     return (
-      <div className='rs-detail-section'>
```
akrikos: Odd that this wasn't using the DetailsSection component already. :+1:

src/DetailsSection.jsx
```diff
@@ -1,11 +1,31 @@
...
+    super(props);
+
+    if (props.isCollapsible === true) {
+      this.state = { changedClass: 'expanded' };
+    } else {
+      this.state = { changedClass: undefined };
+    }
+  }
+
+  collapseSection() {
```
akrikos: This would make more sense if named something like "toggleCollapse"

test/DetailsSectionSpec.jsx
```diff
@@ -13,18 +13,66 @@ describe('DetailsSection', () => {
...
   });
+
+  it('has only default class when custom class not provided', () => {
+    detailsSection = TestUtils.renderIntoDocument(
+      <DetailsSection title="title">
+        Test Detail List
+      </DetailsSection>
+    );
+
+    expect(ReactDOM.findDOMNode(detailsSection)).toHaveClass('rs-detail-section rs-collapsible-section');
```
akrikos: It doesn't make sense to me that the class `rs-collapsible-section` is applied to when a DetailsSection is not collapsible. That class should only be applied when the details section is collapsible. Or am I missing something?

test/DetailsSectionSpec.jsx
```diff
@@ -13,18 +13,66 @@ describe('DetailsSection', () => {
...
 
     detailsSection = TestUtils.renderIntoDocument(
-      <DetailsSection><span className="test-child">test value</span></DetailsSection>
+      <DetailsSection title="title"><span className="test-child">test value</span></DetailsSection>
     );
     sectionChild = TestUtils.findRenderedDOMComponentWithClass(detailsSection, 'test-child');
     expect(sectionChild.getDOMNode().textContent).toBe('test value');
   });
+
+  it('has only default class when custom class not provided', () => {
```
akrikos: Please also test that the caret is not rendered when isCollapsible is false.

src/DetailsSection.jsx
```diff
@@ -1,11 +1,31 @@
...
+      this.setState({ changedClass: 'collapsed' });
+    } else if (this.state.changedClass === 'collapsed') {
+      this.setState({ changedClass: 'expanded' });
+    }
+  }
+
   render() {
     return (
-      <div className="rs-detail-section">
+      <div isCollapsible={ this.props.isCollapsible } onClick={ this.collapseSection.bind(this) } className={ classnames('rs-detail-section rs-collapsible-section', this.state.changedClass) }>
```
akrikos: Do you need the `isCollapsible=...` attribute here? It doesn't seem like it would do anything.

willcumbie: What does an isCollapsible attribute on a div do? Is there any reason to have this here?

test/DetailsSectionSpec.jsx
lucasrose: I agree with you on that, however thats how it looked like on many non-collapsible details sections, like this one in security groups
screen shot 2016-06-20 at 9 39 42 am

src/DetailsSection.jsx
willcumbie: Isn't this going to collapse the section even if it isn't collapsible?

src/DetailsSection.jsx
```diff
@@ -1,11 +1,31 @@
 import React from 'react';
+import classnames from 'classnames';
 
 export default class DetailsSection extends React.Component {
+  constructor(props) {
+    super(props);
+
+    if (props.isCollapsible === true) {
+      this.state = { changedClass: 'expanded' };
```
akrikos: `changedClass` should be named something a bit more self-explanatory to reduce cognitive load when maintaining this component later. When I think about it, it should probably just be a boolean `isCollapsed` and then you should translate that into the correct className/lack of className in the render method.

As is, if you store the actual className like this, it works, but you're mixing the "how" of making the item collapsed with the logical "what" of 'should it be collapsed'. In this case I think mixing the two makes the component a touch more difficult to understand.

test/DetailsSectionSpec.jsx
willcumbie: That's bad. We shouldn't copy that mistake into react-canon.

src/DetailsSection.jsx
lucasrose: @willcumbie no it will not because if it is undefined (meaning it is not a collapsible section) then the state will never change

akrikos:
Looks pretty good. I made some comments on how to help tighten it up a bit. Let me know if any of the comments are unclear.

test/DetailsSectionSpec.jsx
```diff
@@ -28,29 +28,32 @@ describe('DetailsSection', () => {
...
+  it('has only default class when custom class not provided and does not render caret', () => {
+    let caretRendered;
     detailsSection = TestUtils.renderIntoDocument(
       <DetailsSection title="title">
         Test Detail List
       </DetailsSection>
     );
 
-    expect(ReactDOM.findDOMNode(detailsSection)).toHaveClass('rs-detail-section rs-collapsible-section');
+    caretRendered = TestUtils.scryRenderedDOMComponentsWithClass(detailsSection, 'rs-caret').length !== 0;
```
lucasrose: renaming to caret rendered and changing the test made it more readable

test/DetailsSectionSpec.jsx
```diff
@@ -28,29 +28,32 @@ describe('DetailsSection', () => {
...
+    let caretRendered;
     detailsSection = TestUtils.renderIntoDocument(
       <DetailsSection title="title" isCollapsible={ true }>
         Test Detail List
       </DetailsSection>
     );
-    caret = TestUtils.findRenderedDOMComponentWithClass(detailsSection, 'rs-caret');
-    console.log(caret);
+
+    caretRendered = TestUtils.scryRenderedDOMComponentsWithClass(detailsSection, 'rs-caret').length === 1;
```
lucasrose: wanted to be consistent with naming in tests, and this felt more readable

src/DetailsSection.jsx
```diff
@@ -6,28 +6,36 @@ export default class DetailsSection extends React.Component {
...
-    } else if (this.state.changedClass === 'collapsed') {
-      this.setState({ changedClass: 'expanded' });
+  toggleCollapse() {
+    if (this.props.isCollapsible) {
+      this.setState({isCollapsed: !this.state.isCollapsed});
     }
   }
 
   render() {
+    var collapsedClassName, showCaret;
```
lucasrose: this added more code, but logically makes more sense now. Instead of having 3 states and like @akrikos and @willcumbie mentioned, there won't be a possibility of switching into a weird state

willcumbie:
It'd be good to also test that clicking a non-collapsible section doesn't do anything to expand/collapse it. Other than that, LGTM.

akrikos:
Looks good. :+1: No more :cow::bell: needed (have you seen that SNL?).

If you want to be extra-thorough, you could also test that clicking twice toggles one way, then the other. There's no pressing need to do that though.

lucasrose:
@akrikos added thorough tests 👍

adampickeral:
we often render collapsible sections as collapsed by default. this section should have a `defaultCollapsed` prop (or some name like that) so a user can collapse it by default.

adampickeral:
LGTM

PR #109, opened 2016-06-17
Add list table

wpwood:
# List Table

Tables are described at http://rackerlabs.github.io/canon/ui-components/#tables. A working table example can be found in DemoListTableSection.jsx.

screen shot 2016-07-15 at 11 34 52 am

screen shot 2016-07-15 at 11 35 37 am

List tables and embedded list tables are similar, except for the component wrapping them. The regular list table is not wrapped in a border, while the embedded list table has a fixed size and is wrapped in a border.

## Wrapper components

The main component for a list table is `ListTable`, and the main component for the embedded list table is `EmbeddedListTable`. Both of these components take the following properties:

- `overlayStatus` - takes a value of 'none', 'empty', 'error', or 'loading', and describes what overlay should be shown on the table.
- `emptyOverlay` - an `EmptyOverlay` component describing the table's empty state.
- `errorOverlay` - an `ErrorOverlay` component describing the table's error state.
- `loadingOverlay` - a `LoadingOverlay` component describing the table's loading state.

The `EmbeddedListTable` takes an additional property, `size`, which can take the values 'small', 'medium', and 'large', and which specifies how tall the embedded list table will be. These correspond to the small, medium, and large sizes described in http://rackerlabs.github.io/canon/ui-components/#tables for the embedded list table.

In addition, there is an `OverlayStatus` enum that contains the values `NONE`, `EMPTY`, `ERROR`, and `LOADING`, and which can be used for the `overlayStatus` values.  

## ListTableHeader

The `ListTableHeader` component specifies how the table header is rendered, as well as how sorting is handled. The children of the `ListTableHeader` are assumed to be `th` components or a component that will resolve to a `th` component that describes a specific header cell. For instance, a valid `ListTableHeader` is:

```
<ListTableHeader>
  <th className="rs-table-status" />
  <th className="rs-table-sort-text">Name</th>
</ListTableHeader>
``` 

There are several predefined header cell components that implement behavior that conforms to the `th` behavior described for canon-react. These are:

- `CheckboxColumnHeader`
- `StatusColumnHeader`
- `TextColumnHeader` - implements sortable behavior, calls the onSort callback when clicked, displays the sort direction indicator.

When a collection of objects is passed to the list table, it is assumed to be in the order that the user of the component wants it in. The `ListTableHeader` takes the following properties for handling sorting:

- `onSort` - a callback function that is called with the current sort direction and columnId for a sortable header. Any header that has the `sortable` boolean property will be sortable, although the `TextColumnHeader` is the only predefined header that supports that property out of the box. If the header is sortable, it must also support the `columnId` property.
- `sortColumn` - the `columnId` for the column that the collection is sorted by.
- `direction` - current sort direction, either 'asc' or 'desc'.  There is a `SortDirection` enum with values `ASCENDING` and `DESCENDING`, which may be used for the direction value.

## ListTableBody

The `ListTableBody` component takes a collection of data, and then for each object in the collection, it clones its children and passes the object as the child's `instance` property.

```
const collection = [
  {id: 1, name: 'one', count: 1},
  {id: 2, name: 'two', count: 42},
  {id: 3, name: 'three', count: 12},
  {id: 4, name: 'four', count: 6}
];

<ListTableBody collection={ collection }>
  <ListTableRow data-model_id={ instance.id } instance={ instance }>
    <td className="rs-table-status rs-table-status-ok"></td>
    <td className="rs-table-checkbox"><input type="checkbox" /></td>
    <td className="rs-table-text">{ instance.name }</td>
    <td className="rs-table-text">{ instance.count }</td>
  </ListTableRow>
</ListTableBody>
```

The `ListTableRow` component is a convenience for passing the `instance` property to each of its children, but in this case, one could just as easily have used a `tr`. The `ListTableBody` will also set the `key` property for the cloned children. By default, this will be the `id` of the instance. If the instance does not have a unique `id` property, then another unique property can be supplied by setting the `keyGenerator` property on the `ListTableBody`.

# Overlays

There are three predefined Overlay components, the `EmptyOverlay`, the `ErrorOverlay`, and the `LoadingOverlay`. Each of them is intended as a default overlay for the `ListTable` and `EmbeddedListTable` wrapper components. There is also a base `ListTableOverlay` which each of these predefined overlays uses, and which may be used if the table needs some non-default overlay behavior.

screen shot 2016-07-15 at 11 35 06 am
screen shot 2016-07-15 at 11 35 19 am
screen shot 2016-07-15 at 11 35 28 am

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -67,6 +67,7 @@ module.exports = function (grunt) {
           'test_bundle.js': ['test-built/**/*.js']
         },
         options: {
+          external: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment'],
           verbose: true
         }
       },
@@ -75,6 +76,7 @@ module.exports = function (grunt) {
           'canon-react.js': ['transpiled/index.js']
         },
         options: {
+          external: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment'],
           transform: [ 'browserify-shim' ],
           browserifyOptions: {
             standalone: 'canonReact'
@@ -86,6 +88,7 @@ module.exports = function (grunt) {
           'demo/bundle.js': ['transpiled/**/*.js'],
         },
         options: {
+          external: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment'],
           transform: [ 'browserify-shim' ]
         }
       }
@@ -130,7 +133,9 @@ module.exports = function (grunt) {
     'lint:eslint',
     'babel:src',
     'babel:test',
+    'babel:demo',
     'browserify:test',
+    'browserify:demo',
     'browserify:release',
     'uglify:build'
   ]);
--- a/demo/DemoListTableSection.jsx
+++ b/demo/DemoListTableSection.jsx
@@ -1,0 +1,167 @@
+import React from 'react';
+
+import CheckboxColumnHeader from './CheckboxColumnHeader';
+import EmbeddedListTable from './EmbeddedListTable';
+import EmptyOverlay from './EmptyOverlay';
+import ErrorOverlay from './ErrorOverlay';
+import ListTable from './ListTable';
+import ListTableBody from './ListTableBody';
+import ListTableHeader from './ListTableHeader';
+import ListTableRow from './ListTableRow';
+import LoadingOverlay from './LoadingOverlay';
+import OverlayStatus from './OverlayStatus';
+import SortDirection from './SortDirection';
+import StatusColumnHeader from './StatusColumnHeader';
+import TextColumnHeader from './TextColumnHeader';
+
+const DemoListTableHeader = ({ onSort, sortColumn, direction }) => (
+  <ListTableHeader { ...{ onSort, sortColumn, direction } }>
+    <StatusColumnHeader />
+    <CheckboxColumnHeader />
+    <TextColumnHeader columnId="name" className="name" sortable label="Name" />
+    <TextColumnHeader columnId="count" className="count" sortable label="Count" />
+  </ListTableHeader>
+);
+
+DemoListTableHeader.propTypes = {
+  onSort: React.PropTypes.func,
+  sortColumn: React.PropTypes.string,
+  direction: React.PropTypes.oneOf(Object.keys(SortDirection))
+};
+
+const DemoListTableRow = ({ instance }) => (
+  <ListTableRow key={ instance.id } data-model_id={ instance.id } instance={ instance }>
+    <td className="rs-table-status rs-table-status-ok"></td>
+    <td className="rs-table-checkbox"><input type="checkbox" /></td>
+    <td className="rs-table-text">{ instance.name }</td>
+    <td className="rs-table-text">{ instance.count }</td>
+  </ListTableRow>
+);
+
+DemoListTableRow.propTypes = {
+  instance: React.PropTypes.object.isRequired
+};
+
+class DemoListTableSection extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      sortColumn: 'name',
+      direction: SortDirection.ASCENDING
+    };
+
+    this._handleSort = this._handleSort.bind(this);
+  }
+
+  render() {
+    const sortedCollection = this._sort();
+
+    const tableStyle = {width: '500px'};
+
+    return (
+      <div className='rs-detail-section'>
+        <div className='rs-detail-section-header'>
+          <h2>List Table</h2>
+        </div>
+        <div className='rs-detail-section-body'>
+          <h3>List Table</h3>
+          <ListTable style={ tableStyle }>
+            <DemoListTableHeader onSort={ this._handleSort } sortColumn={ this.state.sortColumn } direction={ this.state.direction } />
+            <ListTableBody collection={ sortedCollection }>
+              <DemoListTableRow />
+            </ListTableBody>
+          </ListTable>
+          <hr />
+          <h3>Empty Overlay</h3>
+          <ListTable style={ tableStyle } overlayStatus={ OverlayStatus.EMPTY }>
+            <DemoListTableHeader onSort={ this._handleSort } sortColumn={ this.state.sortColumn } direction={ this.state.direction } />
+            <ListTableBody collection={ [] }>
+              <DemoListTableRow />
+            </ListTableBody>
+          </ListTable>
+          <hr />
+          <h3>Error Overlay</h3>
+          <ListTable style={ tableStyle } overlayStatus={ OverlayStatus.ERROR }>
+            <DemoListTableHeader onSort={ this._handleSort } sortColumn={ this.state.sortColumn } direction={ this.state.direction } />
+            <ListTableBody collection={ [] }>
+              <DemoListTableRow />
+            </ListTableBody>
+          </ListTable>
+          <hr />
+          <h3>Loading Overlay</h3>
+          <ListTable style={ tableStyle } overlayStatus={ OverlayStatus.LOADING }>
+            <DemoListTableHeader onSort={ this._handleSort } sortColumn={ this.state.sortColumn } direction={ this.state.direction } />
+            <ListTableBody collection={ [] }>
+              <DemoListTableRow />
+            </ListTableBody>
+          </ListTable>
+          <hr />
+          <h3>Embedded List Table</h3>
+          <EmbeddedListTable size="medium" style={ tableStyle }>
+            <DemoListTableHeader onSort={ this._handleSort } sortColumn={ this.state.sortColumn } direction={ this.state.direction } />
+            <ListTableBody collection={ sortedCollection }>
+              <DemoListTableRow />
+            </ListTableBody>
+          </EmbeddedListTable>
+        </div>
+      </div>
+    );
+  }
+
+  _handleSort(sortDirection, columnId) {
+    this.setState({
+      sortColumn: columnId,
+      direction: sortDirection
+    });
+  }
+
+  _sort() {
+    return Array.from(COLLECTION).sort(this._compareRows.bind(this));
+  }
+
+  _compareRows(row1, row2) {
+    if (this.state.sortColumn === 'name') {
+      return this._compareStringValues(row1.name, row2.name);
+    } else if (this.state.sortColumn === 'count') {
+      return this._compareNumericValues(row1.count, row2.count);
+    }
+  }
+
+  _compareStringValues(a, b) {
+    if (this.state.direction === SortDirection.ASCENDING) {
+      return a.localeCompare(b);
+    } else {
+      return b.localeCompare(a);
+    }
+  }
+
+  _compareNumericValues(a, b) {
+    if (a === b) {
+      return 0;
+    }
+
+    if (this.state.direction === SortDirection.ASCENDING) {
+      if (a < b) {
+        return -1;
+      } else {
+        return 1;
+      }
+    } else {
+      if (a > b) {
+        return -1;
+      } else {
+        return 1;
+      }
+    }
+  }
+}
+
+const COLLECTION = [
+  {id: 1, name: 'one', count: 1},
+  {id: 2, name: 'two', count: 42},
+  {id: 3, name: 'three', count: 12},
+  {id: 4, name: 'four', count: 6}
+];
+
+export default DemoListTableSection;
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "babelify": "7.2.0",
     "browserify": "<4.0.0",
     "browserify-shim": "3.8.12",
+    "enzyme": "2.3.0",
     "eslint": "^1.10.3",
     "eslint-plugin-react": "^3.15.0",
     "grunt": "~0.4.2",
--- a/src/ListTable.jsx
+++ b/src/ListTable.jsx
@@ -1,0 +1,45 @@
+import { Component, PropTypes } from 'react';
+import ListTableOverlaySelector from './ListTableOverlaySelector';
+import EmptyOverlay from './EmptyOverlay';
+import ErrorOverlay from './ErrorOverlay';
+import LoadingOverlay from './LoadingOverlay';
+import OverlayStatus from './OverlayStatus';
+
+class ListTable extends Component {
+  render() {
+    const {
+      children,
+      emptyOverlay,
+      errorOverlay,
+      loadingOverlay,
+      overlayStatus
+    } = this.props;
+
+    return (
+      <div { ...this.props } >
+        <table className="rs-list-table">
+          { children }
+        </table>
+        <ListTableOverlaySelector { ...{ overlayStatus, emptyOverlay, errorOverlay, loadingOverlay } } />
+      </div>
+    );
+  }
+}
+
+ListTable.defaultProps = {
+  emptyOverlay: <EmptyOverlay title="Empty List" subtitle="This list is empty" message="There is no data to display" />,
+  errorOverlay: <ErrorOverlay message="There was an error loading data" />,
+  loadingOverlay: <LoadingOverlay />,
+  overlayStatus: OverlayStatus.NONE
+};
+
+ListTable.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  emptyOverlay: PropTypes.node,
+  errorOverlay: PropTypes.node,
+  loadingOverlay: PropTypes.node,
+  overlayStatus: PropTypes.oneOf(Object.keys(OverlayStatus))
+};
+
+export default ListTable;
--- a/src/ListTableBody.jsx
+++ b/src/ListTableBody.jsx
@@ -1,0 +1,40 @@
+import { Component, PropTypes } from 'react';
+
+class ListTableBody extends Component {
+  constructor() {
+    super();
+
+    this._cloneRow = this._cloneRow.bind(this);
+  }
+
+  render() {
+    return (
+      <tbody>
+        { this._cloneRowPerInstance() }
+      </tbody>
+    );
+  }
+
+  _cloneRowPerInstance() {
+    return this.props.collection.map(this._cloneRow);
+  }
+
+  _cloneRow(instance, index, array) {
+    return React.cloneElement(this.props.children, {
+      instance: instance,
+      key: this.props.keyGenerator(instance, index, array)
+    });
+  }
+}
+
+ListTableBody.defaultProps = {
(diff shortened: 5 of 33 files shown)


Jigarsolanki:
Could you add some description and usage of it?

src/DataTable.jsx
```diff
@@ -0,0 +1,45 @@
+import { Component, PropTypes } from 'react';
+
+class DataTable extends Component {
```
Jigarsolanki: Why do we call it `DataTable`? why not just `Table`?

src/DataTable.jsx
```diff
@@ -0,0 +1,45 @@
...
+    } = this.props;
+
+    return (
+      <div { ...this.props } >
+        <table className="rs-list-table">
+          { children }
+        </table>
+        { isEmpty ? emptyOverlay : null }
+        { isLoading ? loadingOverlay : null }
+        { isErrored ? errorOverlay : null }
```
MACSkeptic: if all three are true (empty, loading and errored), do we render three overlays on top of each other?

should the status prop be a enum instead?

src/DataTable.jsx
```diff
@@ -0,0 +1,45 @@
...
+DataTable.defaultProps = {
+  isEmpty: false,
+  isErrored: false,
+  isLoading: false
+};
+
+DataTable.propTypes = {
+  children: PropTypes.node,
+  className: React.PropTypes.string,
+  emptyOverlay: PropTypes.node,
```
MACSkeptic: should we make these child components with a specific type instead of being props?

src/DataTableBody.jsx
```diff
@@ -0,0 +1,29 @@
...
+  render() {
+    return (
+      <tbody>
+        { this._cloneRowPerInstance() }
+      </tbody>
+    );
+  }
+
+  _cloneRowPerInstance() {
+    return this.props.collection.map(this._cloneRow.bind(this));
```
jdrush89: bind in the constructor

demo/DemoPopoverSection.jsx
```diff
@@ -104,7 +104,7 @@ class DemoPopoverSection extends React.Component {
                   <span>
                     To top align the content on a submittable popover declare the 'horizontal' prop as false:
                   </span>
-                  <br>
+                  <br />
```
hannahstwinter: 😅 Thanks!

Gruntfile.js
```diff
@@ -67,6 +67,7 @@ module.exports = function (grunt) {
           'test_bundle.js': ['test-built/**/*.js']
         },
         options: {
+          exclude: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment'],
```
MACSkeptic: for my benefit: why are those needed?

src/DataTableBody.jsx
```diff
@@ -0,0 +1,29 @@
...
+  _cloneRow(instance) {
+    return React.cloneElement(this.props.children, {
+      instance: instance,
+      key: instance.id
+    });
+  }
+}
+
+DataTableBody.propTypes = {
+  children: PropTypes.element.isRequired,
```
hannahstwinter: I wonder if this should be a TableRow component instead, implemented with loads of flexibility in terms of structuring. That way it offers a bit of documentation as to intended usage.

Gruntfile.js
wpwood: They cause build errors when updating enzyme if they're not there. Actually, it looks like I ned to change this, according to https://github.com/airbnb/enzyme/blob/master/docs/guides/webpack.md

src/DataTable.jsx
hannahstwinter: I agree this could use more consideration... A status enum might not be a bad thing, that way you can control what values are passed in and possibly even offer default overlay props so it both documents itself and still works without having to touch all of the parts. Also... will React choke on rendering undefined values? If someone passes in isEmpty as true but no emptyOverlay, what will happen? I like trying to code these with the assumption that anyone using it is going to try to break it. 😜

src/DataTableBody.jsx
```diff
@@ -0,0 +1,29 @@
...
+class DataTableBody extends Component {
+  render() {
+    return (
+      <tbody>
+        { this._cloneRowPerInstance() }
+      </tbody>
+    );
+  }
+
+  _cloneRowPerInstance() {
```
MACSkeptic: 👍  clever

I wonder if there are scenarios where this wouldn't be desirable (like certain rows rendering differently enough that you'd want different components)

I guess that could be handled within the row itself

src/EmbeddedDataTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+    const sizeClass = SIZE_CLASSES[size];
+
+    classes = classNames(
+      'rs-embedded-list-table-wrapper',
+      sizeClass,
+      className
+    );
+
+    return (
+      <div { ...this.props } className={ classes }>
```
hannahstwinter: Why do you need ```{ ...this.props }``` here?

src/DataTableBody.jsx
```diff
@@ -0,0 +1,29 @@
...
+    return this.props.collection.map(this._cloneRow.bind(this));
+  }
+
+  _cloneRow(instance) {
+    return React.cloneElement(this.props.children, {
+      instance: instance,
+      key: instance.id
```
MACSkeptic: is "id" always a safe key?

(asking that after I've created quite a few lists with `awsAccountNumber` or something else as the key)

of course we could just map the input to match, or should we accept a function to generate the key?

Gruntfile.js
wpwood: Oh, we're not using webpack, so I'll mark them as externals in browserify as shown in http://benclinkinbeard.com/posts/external-bundles-for-faster-browserify-builds/

The actual build error is `Error: Cannot find module 'react/lib/ReactContext' from '/Users/bill.woodward/Projects/canon-react/node_modules/enzyme/build'`

src/DataTableHeader.jsx
```diff
@@ -0,0 +1,49 @@
...
+      onSort(direction, child.props.columnId);
+    };
+  }
+}
+
+DataTableHeader.defaultProps = {
+  onSort: () => {}
+};
+
+DataTableHeader.propTypes = {
```
MACSkeptic: are all tables sortable? maybe I missed where one could opt out

src/DataTableBody.jsx
MACSkeptic: 👍  big fan of making it a child component instead

Gruntfile.js
MACSkeptic: 👍

demo/DemoPopoverSection.jsx
MACSkeptic: should we build the demo as part of CI?

src/DataTable.jsx
MACSkeptic: or `ListTable` if we're following canon terminology

demo/DemoPopoverSection.jsx
wpwood: Maybe so. I found this when running `scripts/server` to run the demo.

src/DataTable.jsx
wpwood: I like using something other than just Table, since it could get confusing differentiating between a Table component and a table DOM element. I'm good with ListTable. I went with DataTable because it really does expect a collection of data (the whole cloning of children thing).

demo/DemoPopoverSection.jsx
wpwood: @adampickeral What do you think? I'm inclined _not_ to build the demo as part of the CI build, since that's building things we ship, and we don't ship the demo.

MACSkeptic: I'm inclined to say we should build it as part of CI

these small mistakes seem very easy to make and go unnoticed, don't think any amount of code review would have caught it

src/DataTable.jsx
MACSkeptic: ^ I share your concern against just calling it Table, `Table` vs `table` is a very nuance distinction and might not be obvious amidst a more complicate component that uses this

demo/DemoPopoverSection.jsx
adampickeral: build it as part of CI

src/DataTable.jsx
wpwood: My concern with passing them as children is two-fold. First, I like the idea of just passing the children of ListTable to the `table`, and if the overlays were also children, we'd have to filter them out somehow. Second, we can't depend on the loading, error, and empty overlays to always be components of type `LoadingOverlay`, `ErrorOverlay`, and `EmptyOverlay`, since the user of the class may want to implement specialized behavior in their own overlay component and pass it in.

MACSkeptic: the overlays could be just declarative wrappers (i.e.: just a div with the right classes with children) for the filtering

src/EmbeddedDataTable.jsx
wpwood: I may not, but we typically pass props through to the rendered component.

src/DataTable.jsx
wpwood: I guess I'm not clear what that would look like. The whole point of wrapping behavior in a component is to abstract it out so the caller doesn't need to know what's in it. It sounds like you're suggesting not doing that, in order to keep from passing in components as props, which we do quite a bit in canon-react. I don't always love it, but I think it makes sense here.

Can you show me an example of what this would look like?

MACSkeptic: it would be a lot more code for this definitely... something like:

```javascript

let loadingOverlay = React.Children.toArray(this.props.children).find(c => c.type === LoadingOverlay);
let errorOverlay = React.Children.toArray(this.props.children).find(c => c.type === ErrorOverlay);
let emptyOverlay = React.Children.toArray(this.props.children).find(c => c.type === EmptyOverlay);
```

that would allow us to do something like this:

```javascript
<DataTable>
  <LoadingOverlay isLoading={true}>...</LoadingOverlay>
  <ErrorOverlay isErrored={true}>...</ErrorOverlay>
  <EmptyOverlay isEmpty={true}>...</EmptyOverlay>
  <DataTableBody>...</DataTableBody>
</DataTable>
```

I think it is totally debatable which one is better, I'm fine if we say we just go with props : )

demo/DemoListTableSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  _compareNumericValues(a, b) {
+    if (a === b) {
+      return 0;
+    }
+
+    if (this.state.direction === SortDirection.ASCENDING) {
+      if (a < b) {
```
MACSkeptic: `return (a - b) / Math.abs(a - b);` ?

demo/DemoListTableSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+
+  _compareStringValues(a, b) {
+    if (this.state.direction === SortDirection.ASCENDING) {
+      return a.localeCompare(b);
+    } else {
+      return b.localeCompare(a);
+    }
+  }
+
+  _compareNumericValues(a, b) {
```
MACSkeptic: would this work for numeric values that are `'1', '2', '3'` instead of `1, 2, 3` (i.e.: numeric values, but as strings) ?

wpwood: Probably not. This is just the demo, so I didn't feel like it needed any extra bells and whistles.

MACSkeptic: didn't realise this was the demo 😊

MACSkeptic:
:shipit:

adampickeral:
would you mind adding some documentation about the public components (props needed, purpose, etc) being added and how they are used to the description of the PR?

wpwood:
@adampickeral Do you have an example of what you'd like to see? It seems like the documentation we have is primarily the Demo classes, which I've provided. Also, it probably shouldn't reside in the PR, but the wiki is currently empty, so opinions on where it should go?

adampickeral:
when we have a larger feature being added such as this, it's good to have documentation in the PR (or in an issue proposing the change) about the approach that is being taken. It makes reviewing the approach at a higher level easier, and it also makes it easier for anyone to update the official documentation after the feature is merged in.

https://github.com/rackerlabs/canon-react/pull/27
https://github.com/rackerlabs/canon-react/issues/55

wpwood:
@adampickeral Gotcha. Thanks for the examples. I'll add something similar.

adampickeral:
:+1:

package.json
```diff
@@ -53,6 +53,7 @@
     "babelify": "7.2.0",
     "browserify": "<4.0.0",
     "browserify-shim": "3.8.12",
+    "enzyme": "2.3.0",
```
willcumbie: :+1:

src/ListTableBody.jsx
```diff
@@ -0,0 +1,40 @@
...
+        { this._cloneRowPerInstance() }
+      </tbody>
+    );
+  }
+
+  _cloneRowPerInstance() {
+    return this.props.collection.map(this._cloneRow);
+  }
+
+  _cloneRow(instance) {
```
willcumbie: map() passes three arguments--the current item, the index, and the array as a whole. We should at least pass the first two into the key generator, it's not hard to imagine a scenario where the index might be needed there.

willcumbie:
LGTM

demo/DemoListTableSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+    const sortedCollection = this._sort();
+
+    const tableStyle = {width: '500px'};
+
+    const emptyOverlay = (<EmptyOverlay title="Empty Overlay" subtitle="I am so empty" message="Empty, empty, empty" />);
+    const errorOverlay = (<ErrorOverlay message="There was an error loading data" />);
+    const loadingOverlay = (<LoadingOverlay />);
+
+    return (
+      <div className='rs-detail-section'>
```
hannahstwinter: Should we use Canon-React components in the demo? It would suck if something were updated/modified at some point and the demo broke.

src/ListTable.jsx
```diff
@@ -0,0 +1,39 @@
...
+      loadingOverlay,
+      overlayStatus
+    } = this.props;
+
+    return (
+      <div { ...this.props } >
+        <table className="rs-list-table">
+          { children }
+        </table>
+        <ListTableOverlaySelector { ...{overlayStatus, emptyOverlay, errorOverlay, loadingOverlay} } />
```
hannahstwinter: spacing nit - ```{ ...{ overlayStatus, emptyOverlay, errorOverlay, loadingOverlay } }```

src/TextColumnHeader.jsx
```diff
@@ -0,0 +1,67 @@
...
+  label: '',
+  sortable: false,
+  onSort: () => {}
+};
+
+TextColumnHeader.propTypes = {
+  label: PropTypes.string.isRequired,
+  sortable: PropTypes.bool.isRequired,
+  onSort: PropTypes.func.isRequired,
+  direction: PropTypes.oneOf([
```
hannahstwinter: PropTypes.oneOf(Object.keys(SortDirection))

demo/DemoListTableSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+    <CheckboxColumnHeader />
+    <TextColumnHeader columnId="name" className="name" sortable label="Name" />
+    <TextColumnHeader columnId="count" className="count" sortable label="Count" />
+  </ListTableHeader>
+);
+
+DemoListTableHeader.propTypes = {
+  onSort: React.PropTypes.func,
+  sortColumn: React.PropTypes.string,
+  direction: React.PropTypes.oneOf([
```
hannahstwinter: PropTypes.oneOf(Object.keys(SortDirection))

src/ListTableHeader.jsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+ListTableHeader.defaultProps = {
+  onSort: () => {}
+};
+
+ListTableHeader.propTypes = {
+  onSort: PropTypes.func,
+  sortColumn: PropTypes.string,
+  direction: PropTypes.oneOf([
```
hannahstwinter: PropTypes.oneOf(Object.keys(SortDirection))

src/ListTableRow.jsx
```diff
@@ -0,0 +1,23 @@
+import { Component, PropTypes } from 'react';
+
+class ListTableRow extends Component {
```
hannahstwinter: If this is stateful solely to handle the cloneChildren function as a class function I feel like it can be stored in a var just as easily. I know you </3 stateless components but I do think we should aim for consistency with when we choose to go which route. Refs seem like a reasonable delineator, and structuring logic and flow via class functions makes sense to me. This wouldn't convolute the latter, IMO, and doesn't touch the first point. Just a thought.

src/TextColumnHeader.jsx
```diff
@@ -0,0 +1,67 @@
...
+class TextColumnHeader extends Component {
+  render() {
+    let classes;
+    const { label, sortable, direction, className } = this.props;
+
+    if (sortable) {
+      classes = classNames(
+        'rs-table-sort',
+        { 'rs-table-sort-asc': direction === SortDirection.ASCENDING },
+        { 'rs-table-sort-desc': direction === SortDirection.DESCENDING },
```
hannahstwinter: you can make these one obj if you want -
```
{ 
  'rs-table-sort-asc': direction === SortDirection.ASCENDING,
  'rs-table-sort-desc': direction === SortDirection.DESCENDING 
}
```

src/TextColumnHeader.jsx
```diff
@@ -0,0 +1,67 @@
...
+      </th>
+    );
+  }
+
+  _handleClick(ev) {
+    let newDirection;
+    const { direction } = this.props;
+
+    if (direction === SortDirection.ASCENDING) {
+      newDirection = SortDirection.DESCENDING;
```
hannahstwinter: You could call this.props.onSort(SortDir.DESC) inline here and on 42 and save the var. It would read a bit better, IMO, less to track in my brain space - if ascending, sort the other way.

hannahstwinter:
Great work, @wpwood ! This looks awesome. :+1:

demo/DemoListTableSection.jsx
wpwood: Maybe so. It's like this in every demo that uses a section. I'll update it in a separate PR.

PR #111, opened 2016-06-20
feat(details-section):allow collapsibility

lucasrose:
add initialCollapsed, collapsible, id, hasLoaded props

### DIFF
--- a/src/DetailsSection.jsx
+++ b/src/DetailsSection.jsx
@@ -1,39 +1,39 @@
 import React from 'react';
 import classnames from 'classnames';
+import DetailsSectionTitle from './DetailsSectionTitle';
 
-export default class DetailsSection extends React.Component {
+class DetailsSection extends React.Component {
   constructor(props) {
     super(props);
 
-    if (props.isCollapsible === true) {
-      this.state = { isCollapsed: props.defaultCollapsed };
-    }
+    this.state = { isCollapsed: props.initialCollapsed };
   }
 
-  toggleCollapse() {
-    if (this.props.isCollapsible) {
-      this.setState({isCollapsed: !this.state.isCollapsed});
+  toggleCollapsed() {
+    if (this.props.collapsible) {
+      let newState = !this.state.isCollapsed;
+      this.setState({ isCollapsed: newState });
+      this.props.onToggleCollapsed(newState);
     }
   }
 
   render() {
-    var collapsedClassName, showCaret;
+    let classNames, { collapsible, isLoading } = this.props, { isCollapsed } = this.state;
 
-    showCaret = false;
-    if (this.props.isCollapsible) {
-      showCaret = true;
-      if (this.state.isCollapsed) {
-        collapsedClassName = 'rs-collapsible-section collapsed';
-      } else {
-        collapsedClassName = 'rs-collapsible-section expanded';
-      }
-    }
+    classNames = {
+      'rs-collapsible-section collapsed': collapsible && isCollapsed,
+      'rs-collapsible-section expanded': collapsible && !isCollapsed,
+      'loading': isLoading
+    };
 
     return (
-      <div onClick={ this.toggleCollapse.bind(this) } className={ classnames('rs-detail-section', collapsedClassName) }>
+      <div { ...this.props } onClick={ this.toggleCollapsed.bind(this) }
+        className={ classnames('rs-detail-section', this.props.className, classNames) }>
         <div className='rs-detail-section-header'>
-          { showCaret ? <div className='rs-caret'></div> : null}
-          <div className='rs-detail-section-title'>{ this.props.title }</div>
+          { this.props.headers }
+          { collapsible ? <div className='rs-caret'></div> : null }
+          { this.props.title ? <DetailsSectionTitle>{ this.props.title }</DetailsSectionTitle> : null}
+          { this.props.subtitle }
         </div>
         <div className='rs-detail-section-body'>
           {this.props.children}
@@ -44,14 +44,22 @@ export default class DetailsSection extends React.Component {
 }
 
 DetailsSection.propTypes = {
-  title: React.PropTypes.node.isRequired,
+  title: React.PropTypes.node,
   children: React.PropTypes.node,
+  headers: React.PropTypes.node,
+  isLoading: React.PropTypes.bool,
+  initialCollapsed: React.PropTypes.bool,
+  onToggleCollapsed: React.PropTypes.func,
+  collapsible: React.PropTypes.bool,
   className: React.PropTypes.string,
-  isCollapsible: React.PropTypes.bool,
-  defaultCollapsed: React.PropTypes.bool
+  subtitle: React.PropTypes.node
 };
 
 DetailsSection.defaultProps = {
-  isCollapsible: false,
-  defaultCollapsed: false
+  collapsible: false,
+  initialCollapsed: false,
+  isLoading: false,
+  onToggleCollapsed: () => {}
 };
+
+export default DetailsSection;
--- a/test/DetailsSectionSpec.jsx
+++ b/test/DetailsSectionSpec.jsx
@@ -1,21 +1,49 @@
 import DetailsSection from '../transpiled/DetailsSection';
+import DetailsSectionSubtitle from '../transpiled/DetailsSectionSubtitle';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
 
 describe('DetailsSection', () => {
-  let detailsSection;
+  let detailsSection, currentState, toggleState;
 
   afterEach(() => {
     ReactDOM.unmountComponentAtNode(ReactDOM.findDOMNode(detailsSection).parentNode);
   });
 
-  it('displays the title', () => {
-    let title;
+  toggleState = ((isCollapsed) => {
+    currentState = isCollapsed;
+  });
+
+  it('displays the title but no subtitle without prop', () => {
+    let title, hasSubtitle;
 
     detailsSection = TestUtils.renderIntoDocument(<DetailsSection title="title" />);
     title = TestUtils.findRenderedDOMComponentWithClass(detailsSection, 'rs-detail-section-title');
+    hasSubtitle = TestUtils.scryRenderedDOMComponentsWithClass(detailsSection, 'rs-detail-section-subtitle').length === 1;
     expect(title.getDOMNode().textContent).toBe('title');
+    expect(hasSubtitle).toBe(false);
+  });
+
+  it('does not display a title when not passed as prop', () => {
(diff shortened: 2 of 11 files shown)


src/DetailsSection.jsx
```diff
@@ -5,22 +5,23 @@ export default class DetailsSection extends React.Component {
...
+    if (props.collapsible) {
+      this.state = { isCollapsed: props.initialCollapsed };
     }
   }
 
-  toggleCollapse() {
-    if (this.props.isCollapsible) {
-      this.setState({isCollapsed: !this.state.isCollapsed});
+  toggleCollapsed() {
+    debugger;
```
willcumbie: :scissors:

src/DetailsSection.jsx
```diff
@@ -29,8 +30,12 @@ export default class DetailsSection extends React.Component {
       }
     }
 
+    if (!this.props.hasLoaded) {
+      loading = 'loading';
+    }
+
     return (
-      <div onClick={ this.toggleCollapse.bind(this) } className={ classnames('rs-detail-section', collapsedClassName) }>
+      <div id={ this.props.id } onClick={ this.toggleCollapsed.bind(this) } className={ classnames('rs-detail-section', collapsedClassName, loading) }>
```
willcumbie: Instead of passing `id` through specifically, how about passing `{...this.props}` here? That way we won't have to update Canon-React each time someone needs to add a new attribute to that section.

src/DetailsSection.jsx
```diff
@@ -46,12 +51,14 @@ export default class DetailsSection extends React.Component {
...
+  collapsible: React.PropTypes.bool,
+  id: React.PropTypes.string
 };
 
 DetailsSection.defaultProps = {
-  isCollapsible: false,
-  defaultCollapsed: false
+  collapsible: false,
+  initialCollapsed: false,
+  hasLoaded: true
```
adampickeral: i think I'd like to keep this consistent with the other props in terms of naming and default behavior:
`isLoading: false`

lucasrose: I only did that because in reach for the details section hasLoaded is used, but I can change it!

src/DetailsSection.jsx
```diff
@@ -29,8 +30,12 @@ export default class DetailsSection extends React.Component {
       }
     }
 
+    if (!this.props.hasLoaded) {
+      loading = 'loading';
+    }
+
```
adampickeral: might be cleaner to pull class names into an object:

```javascript
let classNames, { collapsible, isLoading } = this.props, { isCollapsed } = this.state;

classNames = {
  'rs-collapsible-section collapsed': collapsible && isCollapsed,
  'rs-collapsible-section expanded': collapsible && !isCollapsed
  'loading': isLoading
};
```
then you call the `classnames` function with that object:
```javascript
className={ classnames('rs-detail-section', classNames) }
```

this will also allow you to get rid of the `if` statements above, and you can replace the `showCaret` variable reference in the jsx with just `collapsible`

src/DetailsSection.jsx
akrikos: non-blocking comment (you don't have to make this change)

I have a small preference for `isLoading` (with a default of false) over `hasLoaded`. My reasoning is that `hasLoaded` indicates that something has happened and since the default is `true`, it feels odd since nothing has necessarily occurred. I know that the Reach components did things this way and it is kind of a minor difference.

test/DetailsSectionSpec.jsx
```diff
@@ -86,9 +106,24 @@ describe('DetailsSection', () => {
     expect(caretRendered).toBe(true);
   });
 
+  it('renders collapsible section with hasLoaded false and initially collapsed by initialCollapsed prop', () => {
```
akrikos: Renders it how? This needs to describe what you're testing more clearly. Perhaps `renders section as collapsed when hasLoaded is false...` 

Isn't the section always collapsed when hasLoaded is false?

test/DetailsSectionSpec.jsx
```diff
@@ -60,23 +65,38 @@ describe('DetailsSection', () => {
...
+        Test Detail List
+      </DetailsSection>
+    );
+
+    caretRendered = TestUtils.scryRenderedDOMComponentsWithClass(detailsSection, 'rs-caret').length === 1;
+    expect(ReactDOM.findDOMNode(detailsSection)).toHaveClass('rs-detail-section rs-collapsible-section expanded');
+    expect(caretRendered).toBe(true);
+  });
+
+  it('renders collapsible section with hasLoaded false and caret when passed collapsible true as prop', () => {
```
akrikos: test name needs to more clearly describe what you're testing (same as below)

test/DetailsSectionSpec.jsx
akrikos: Perhaps `renders section as loading and collapsed when...`

src/DetailsSection.jsx
```diff
@@ -29,8 +30,12 @@ export default class DetailsSection extends React.Component {
       }
     }
 
+    if (!this.props.hasLoaded) {
+      loading = 'loading';
```
akrikos: Possibly rename to loadingClassName to be as explicit as you are with the other className variable.

src/DetailsSection.jsx
```diff
@@ -46,12 +40,13 @@ export default class DetailsSection extends React.Component {
 DetailsSection.propTypes = {
   title: React.PropTypes.node.isRequired,
   children: React.PropTypes.node,
-  className: React.PropTypes.string,
-  isCollapsible: React.PropTypes.bool,
-  defaultCollapsed: React.PropTypes.bool
+  isLoading: React.PropTypes.bool,
+  initialCollapsed: React.PropTypes.bool,
+  collapsible: React.PropTypes.bool,
```
willcumbie: Shouldn't have a comma here.

willcumbie:
LGTM

src/DetailsSection.jsx
```diff
@@ -5,34 +5,28 @@ export default class DetailsSection extends React.Component {
...
+      'rs-collapsible-section expanded': collapsible && !isCollapsed,
+      'loading': isLoading
+    };
 
     return (
-      <div onClick={ this.toggleCollapse.bind(this) } className={ classnames('rs-detail-section', collapsedClassName) }>
+      <div { ...this.props } onClick={ this.toggleCollapsed.bind(this) } className={ classnames('rs-detail-section', classNames) }>
         <div className='rs-detail-section-header'>
-          { showCaret ? <div className='rs-caret'></div> : null}
+          { this.props.collapsible ? <div className='rs-caret'></div> : null}
```
adampickeral: just `collapsible`, not `this.props.collapsible`

adampickeral:
you should add a section that is loading in the demo too. also make sure you squash everything into a single commit (everything looks good)

src/DetailsSection.jsx
```diff
@@ -1,39 +1,38 @@
...
   }
 
-  toggleCollapse() {
-    if (this.props.isCollapsible) {
-      this.setState({isCollapsed: !this.state.isCollapsed});
+  toggleCollapsed() {
+    if (this.props.collapsible) {
+      let newState = !this.state.isCollapsed;
+      this.setState({ isCollapsed: newState });
+      this.props.onToggleCollapsed(newState);
```
akrikos: :+1:

src/DetailsSection.jsx
```diff
@@ -1,39 +1,38 @@
...
+      'loading': isLoading
+    };
 
     return (
-      <div onClick={ this.toggleCollapse.bind(this) } className={ classnames('rs-detail-section', collapsedClassName) }>
+      <div { ...this.props } onClick={ this.toggleCollapsed.bind(this) }
+        className={ classnames('rs-detail-section', this.props.className, classNames) }>
         <div className='rs-detail-section-header'>
-          { showCaret ? <div className='rs-caret'></div> : null}
+          { this.props.headers }
```
adampickeral: is this intended to just override they way a title is rendered? looking at the use case on the create server page, it looks like title isn't even passed in, yet we're still rendering the `rs-detail-section-title` element below (and end up with two of them).

perhaps if `headers` is provided, `title` and `subtitle` shouldn't be rendered

src/DetailsSection.jsx
```diff
@@ -46,12 +45,20 @@ export default class DetailsSection extends React.Component {
...
 DetailsSection.defaultProps = {
-  isCollapsible: false,
-  defaultCollapsed: false
+  collapsible: false,
+  initialCollapsed: false,
+  isLoading: false,
+  onToggleCollapsed: () => {}
 };
+
+export default DetailsSection;
```
adampickeral: new line

src/DetailsSection.jsx
lucasrose: what if title was not a required prop, then maybe this instead? `{ this.props.title ? <div className="rs-detail-section-title">{ this.props.title }</div> : null}`  it would take care of when a title is not passed in

adampickeral: yeah that'd work

akrikos:
:+1:

test/DetailsSectionTitleSpec.jsx
```diff
@@ -0,0 +1,39 @@
...
+    expect(ReactDOM.findDOMNode(detailsSectionTitle)).toHaveClass('rs-detail-section-title');
+    expect(ReactDOM.findDOMNode(detailsSectionTitle)).not.toHaveClass('test');
+  });
+
+  it('renders passed in children', () => {
+    detailsSectionTitle = TestUtils.renderIntoDocument(
+      <DetailsSectionTitle>Target Child</DetailsSectionTitle>
+    );
+
+    expect(TestUtils.findRenderedDOMComponentWithClass(detailsSectionTitle, 'rs-detail-section-title')).not.toBeNull();
```
adampickeral: you should be testing that the title component has the right children (`Target Child`) here

test/DetailsSectionSubtitleSpec.jsx
```diff
@@ -0,0 +1,37 @@
...
+    expect(ReactDOM.findDOMNode(detailsSectionSubtitle)).toHaveClass('rs-detail-section-subtitle');
+    expect(ReactDOM.findDOMNode(detailsSectionSubtitle)).not.toHaveClass('test');
+  });
+
+  it('renders passed in children', () => {
+    detailsSectionSubtitle = TestUtils.renderIntoDocument(
+      <DetailsSectionSubtitle>subtitle</DetailsSectionSubtitle>
+    );
+
+    expect(TestUtils.findRenderedDOMComponentWithClass(detailsSectionSubtitle, 'rs-detail-section-subtitle')).not.toBeNull();
```
adampickeral: you should be testing that the subtitle component has the right children (`subtitle`) here

PR #114, opened 2016-06-24
fix form: expose rs-form-horizontal as optional

hannahstwinter:
as the FormPopover component consumes the Form component, exposing the rs-form-horizontal class as optional is critical in the case of top aligned field labels as documented in Canon


willcumbie:
The demo page should also include an example of a non-horizontal form, since it's not obvious when you'd need that.

willcumbie:
:+1:

adampickeral:
:+1:

PR #119, opened 2016-07-06
Fix FormPopover error display [ready]

hannahstwinter:
This PR introduces an ErrorIndicator component to be used in cases such as displaying errors on the FormPopover footer. Currently, the error is being displayed through a FormFieldValidationBlock below the button components. However, the Canon classes being used in that component are meaningless outside of a control group, so the error would display as plain black text and the positioning (being rendered below the buttons) would display the error beside of the buttons, widening the popover substantially.

Current (screenshot of code I'm developing within Reach consuming the FormPopover pattern) -
screen shot 2016-07-06 at 12 20 19 pm

With changes (rendering error through ErrorIndicator component located above buttons) -
screen shot 2016-07-06 at 12 13 13 pm

Canon does not have any patterns surrounding error text outside of tables and validation, so I have settled on the ```rs-status-error``` class here to convey that something is borked and paint it red.
Any thoughts or feedback as to a different approach (including whether or not you think there should be a more appropriate class in the Canon repo) would be appreciated.

### DIFF
--- a/test/FormPopoverSpec.jsx
+++ b/test/FormPopoverSpec.jsx
@@ -4,6 +4,14 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-addons-test-utils';
 
+import Button from '../transpiled/Button';
+import Form from '../transpiled/Form';
+import ErrorIndicator from '../transpiled/ErrorIndicator';
+import PopoverBody from '../transpiled/PopoverBody';
+import PopoverFooter from '../transpiled/PopoverFooter';
+import PopoverOverlay from '../transpiled/PopoverOverlay';
+import ProcessingIndicator from '../transpiled/ProcessingIndicator';
+
 describe('FormPopover', () => {
   let onCancel, onSubmit, tether;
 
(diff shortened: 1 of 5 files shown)


willcumbie:
Screenshot?

jdrush89:
LGTM 🎉

willcumbie:
It'd be better if we had an approved Canon pattern for this kind of thing, but I'm not sure there are even designers working on Canon anymore. This is probably alright in the mean time.

Jigarsolanki:
LGTM

hannahstwinter:
@willcumbie That was my thought on this, too - I was pretty stumped as to which approach to take. Also it did confuse me immensely that we didn't have any sort of basic CSS for errors, esp in popovers (again, they're all very custom to Canon tables or field validation). Ultimately I think a 'for now' fix is ok enough... It feels weird, but less weird than some of the other solutions I came up with. 🙊

test/FormPopoverSpec.jsx
```diff
@@ -21,109 +29,181 @@ describe('FormPopover', () => {
...
+    const shallowRenderPopover = (popoverProps) => {
+      const renderer = TestUtils.createRenderer();
+      renderer.render(formPopover(popoverProps));
+
+      popover = renderer.getRenderOutput();
+      popoverOverlay = popover.props.children;
+      form = popoverOverlay.props.children;
+      popoverBody = form.props.children[0];
+      popoverFooter = form.props.children[1];
+      errorIndicator = popoverFooter.props.children[0];
```
adampickeral: you can save some lines by doing something like this:
```javascript
[popoverBody, popoverFooter] = form.props.children;
[ errorIndicator, submit, cancel, processingIndicator ] = popoverFooter.props.children;
```

test/FormPopoverSpec.jsx
```diff
@@ -21,109 +29,181 @@ describe('FormPopover', () => {
...
 
-    expect(eventCancelSpy).toHaveBeenCalled();
+      expect(onCancel).toHaveBeenCalled();
+    });
   });
 
-  describe('while processing', () => {
+  describe('rendering', () => {
+    let popover, popoverOverlay, form, popoverBody, popoverFooter, errorIndicator,
+        submit, cancel, processingIndicator;
```
adampickeral: so much indentation

hannahstwinter: Noooo! I like it better this way, lining up the names instead... 😢

adampickeral: 2 space indent

adampickeral:
looks good, squash into a single commit

PR #124, opened 2016-07-07
feat(automation): add a script to automatically generate release branches

willcumbie:
Automates the creation of a new release pull request. Follow-up pull request will add a second script to call the GitHub API and create the release with relevant attachments.

### DIFF
--- a/scripts/create_release_branch.py
+++ b/scripts/create_release_branch.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+import re
+import os
+from distutils.version import StrictVersion
+import uuid
+
+os.system("git checkout master")
+os.system("git pull")
+
+# get the initial version number
+node_file = open(os.path.dirname(__file__) + '/../package.json', "r")
+package_json_contents = node_file.readlines()
+node_file.close()
+
+for index, line in enumerate(package_json_contents):
+    version_match = re.search('"version": "([0-9\.]*)"', line)
+    if (version_match != None):
+        version_line = index
+        current_version = version_match.group(1)
+        break
+
+# get the new version number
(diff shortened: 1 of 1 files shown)


Jigarsolanki:
👎 it should be in node.js :) 

LGTM

willcumbie:
Actually thought seriously about trying to do this in Node, but it turns out I hate Node. It'd probably also work as a shell script, though the file manipulation might be a little harder to read.

adampickeral:
it's cool, but i'd prefer to not add python as a dependency to this project

scripts/create_release_branch.py
```diff
@@ -0,0 +1,70 @@
...
+changelog_contents = changelog.readlines()
+changelog.close()
+new_changelog_contents = ['### Version ' + new_version + '\n'] + ['\n'.join(change_lines), '\n\n'] + changelog_contents
+changelog = open(os.path.dirname(__file__) + '/../CHANGELOG.md', "w")
+new_changelog_contents = "".join(new_changelog_contents)
+changelog.write(new_changelog_contents)
+changelog.close()
+
+# commit the branch and push it
+formatted_change_list = '\n'.join(change_lines).replace('"','')
```
akrikos: You may want to warn folks that any double quotes will be removed from the change log on line 31

akrikos:
Other than the nit about warning the user about double quotes (assuming I'm reading that correctly), it looks good.

willcumbie:
@adampickeral I can rewrite this to a shell script, sure, but I feel like that's going to be a poor choice, if not completely unworkable, for the next script, which will need to make API calls and process JSON data. Python is pretty ubiquitous, and it's an optional dependency, no?

adampickeral:
hmm, ok I'll go with it

PR #127, opened 2016-07-08
refactor - migrate simple components to stateless functional components [ready]

hannahstwinter:
Another PR migrating simple components to stateless functional components. There will probably be one more though it will only have two or so components and I'm stalling because the effects are trickling through to other spec files. 🙊

### DIFF
--- a/src/FormFieldValidationBlock.jsx
+++ b/src/FormFieldValidationBlock.jsx
@@ -1,15 +1,15 @@
 import React from 'react';
 
-export default class FormFieldValidationBlock extends React.Component {
-  render() {
-    return this.props.value ? (
-      <span className="rs-validation-block">
-        <i className="rs-validation-indicator"></i> {this.props.value}
-      </span>
-    ) : null;
-  }
-}
+const FormFieldValidationBlock = (props) => {
+  return props.value ? (
+    <span className="rs-validation-block">
+      <i className="rs-validation-indicator"></i> { props.value }
+    </span>
+  ) : <noscript />;
+};
 
 FormFieldValidationBlock.propTypes = {
   value: React.PropTypes.node
 };
+
+export default FormFieldValidationBlock;
--- a/test/FormFieldSpec.jsx
+++ b/test/FormFieldSpec.jsx
@@ -1,72 +1,86 @@
 import FormField from '../transpiled/FormField';
 import React from 'react';
-import ReactDOM from 'react-dom';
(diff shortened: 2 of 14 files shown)


src/FormFieldValidationBlock.jsx
```diff
@@ -1,15 +1,15 @@
...
-      </span>
-    ) : null;
-  }
-}
+const FormFieldValidationBlock = (props) => {
+  return props.value ? (
+    <span className="rs-validation-block">
+      <i className="rs-validation-indicator"></i> { props.value }
+    </span>
+  ) : <noscript />;
```
willcumbie: Why noscript rather than null?

test/FormFieldSpec.jsx
```diff
@@ -1,72 +1,71 @@
...
 
-      message = TestUtils.findRenderedDOMComponentWithClass(formField, 'rs-validation-block');
-      expect(message.getDOMNode().textContent).toBe(' Test success message');
+      expect(validationBlock.props.value).toBe('Test success message');
     });
   });
 
-  it('adds the help message', () => {
-    let message;
+  // it('adds the help message', () => {
```
willcumbie: Uncomment these.

hannahstwinter: wherps

src/FormFieldValidationBlock.jsx
hannahstwinter: Note section on https://facebook.github.io/react/docs/reusable-components.html#stateless-functions - "In React v0.14, stateless functional components were not permitted to return null or false (a workaround is to return a <noscript /> instead). This was fixed in React v15...". I think we're using 0.14.8?

willcumbie:
LGTM

src/FormFieldValidationBlock.jsx
hannahstwinter: lol - it ate my noscript tag

j00k:
:ok_hand:

adampickeral:
:+1: 

for future PR's, commit messages should look a little more like this:
```
refactor(stateless components): migrate components to stateless components
// anything else needed to explain the change in more detail
```
https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format

hannahstwinter:
@adampickeral I rebased and fixed it. I meant to do that before now - thanks for reminding me! I think I forgot to do this on the earlier comparable PR.

PR #128, opened 2016-07-13
fix (FormPopover propType): fix FormPopover target prop declaration [ready]

hannahstwinter:
Previously the target prop on the FormPopover component was defined as being a string. However, the Popover component being rendered by the FormPopover accepts either strings or functions as the target.
This PR changes the FormPopover component to reflect that, preventing the invalid PropType warning from being thrown when the FormPopover is being passed a function as the target (which is a valid means of consuming the Popover and should be reflected in the FormPopover as such).
This is a long explanation for a small change.


willcumbie:
:+1:

jdrush89:
LGTM

adampickeral:
:+1:

PR #129, opened 2016-07-15
feat (FormPopover): add optional additional controls to footer [ready]

hannahstwinter:
Extend FormPopover pattern to optionally render more controls in footer than solely Submit and Cancel. 
This is not an uncommon pattern as popovers can be used as a location to both edit and delete an entity (example from Reach) -
screen shot 2016-07-15 at 11 21 55 am
The popover pictured above exposes both edit functionality as well as disabling/delete functionality.

Screenshot of FormPopover demo -
screen shot 2016-07-15 at 11 21 04 am

### DIFF
--- a/demo/DemoPopoverSection.jsx
+++ b/demo/DemoPopoverSection.jsx
@@ -98,7 +98,8 @@ class DemoPopoverSection extends React.Component {
                     target={ () => document.getElementById('form-popover-button-id') }
                     isOpen={ this.state.formPopoverOpen }
                     onSubmit={ () => { this.setState({ formPopoverSubmitting: true }) } }
-                    onRequestClose={ () => { this.setState({ formPopoverOpen: false }) } } />
+                    onRequestClose={ () => { this.setState({ formPopoverOpen: false }) } }
+                    additionalControls={ [ <Button onClick={ (e) => { e.preventDefault(); } } canonStyle="secondary">Opt. Addl Controls</Button> ] } />
                 </td>
               </tr>
               <tr>
--- a/src/FormPopover.jsx
+++ b/src/FormPopover.jsx
@@ -26,7 +26,7 @@ class FormPopover extends React.Component {
     delete popoverProps.children;
 
     return (
-      <Popover { ...this.props }>
+      <Popover { ...popoverProps }>
         <PopoverOverlay>
           <Form onsubmit={ this.props.onSubmit } size={ this.props.formSize } horizontal={ this.props.horizontal }>
             <PopoverBody>
@@ -35,6 +35,7 @@ class FormPopover extends React.Component {
             <PopoverFooter>
               <ErrorIndicator value={ this.props.error }/>
               { submitComponent }
+              { this.props.additionalControls }
               { cancelComponent }
               <ProcessingIndicator hidden={ !this.props.processing }/>
             </PopoverFooter>
@@ -69,6 +70,7 @@ FormPopover.propTypes = {
   // popover layout
   cancelButton: React.PropTypes.element.isRequired,
   submitButton: React.PropTypes.element.isRequired,
+  additionalControls: React.PropTypes.node,
   // popover content
   children: React.PropTypes.node.isRequired,
   // popover
@@ -86,6 +88,7 @@ FormPopover.defaultProps = {
   horizontal: true,
   cancelButton: <Button canonStyle="link">Cancel</Button>,
   submitButton: <Button canonStyle="primary">Submit</Button>,
+  additionalControls: [],
   processing: false,
   isOpen: true,
   placement: 'right'
--- a/test/FormPopoverSpec.jsx
+++ b/test/FormPopoverSpec.jsx
@@ -87,8 +87,8 @@ describe('FormPopover', () => {
   });
 
   describe('rendering', () => {
-    let popover, popoverOverlay, form, popoverBody, popoverFooter, errorIndicator,
-      submit, cancel, processingIndicator;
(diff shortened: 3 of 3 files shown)


willcumbie:
Hmm. I punted on supporting extra buttons because I figured it would be tricky and likely require a new component; if it's this easy, that's great. I feel like it's a bit of a pain that this new button won't interact with things like onSubmit, disable during processing, etc., but it's probably better that we remain agnostic on those.

Out of curiosity, which popover are you thinking of that needs this?

demo/DemoPopoverSection.jsx
```diff
@@ -98,7 +98,8 @@ class DemoPopoverSection extends React.Component {
                     target={ () => document.getElementById('form-popover-button-id') }
                     isOpen={ this.state.formPopoverOpen }
                     onSubmit={ () => { this.setState({ formPopoverSubmitting: true }) } }
-                    onRequestClose={ () => { this.setState({ formPopoverOpen: false }) } } />
+                    onRequestClose={ () => { this.setState({ formPopoverOpen: false }) } }
+                    additionalControls={ [ <Button onClick={ (e) => { e.preventDefault(); } } canonStyle="secondary">Opt. Addl Controls</Button> ] } />
```
MACSkeptic: doesn't this complain about the lack of `key` ?

test/FormPopoverSpec.jsx
```diff
@@ -164,6 +165,10 @@ describe('FormPopover', () => {
       expect(submit.props.enabled).toBe(true);
     });
 
+    it('rendres the additionalControls if passed in', () => {
```
akrikos: *renders

test/FormPopoverSpec.jsx
```diff
@@ -87,8 +87,8 @@ describe('FormPopover', () => {
   });
 
   describe('rendering', () => {
-    let popover, popoverOverlay, form, popoverBody, popoverFooter, errorIndicator,
-      submit, cancel, processingIndicator;
+    let anotherButton, popover, popoverOverlay, form, popoverBody, popoverFooter,
```
akrikos: This list of variables would be much more readable as one var per line in alphabetical order. Or just in alphabetical order if you prefer.

jdrush89:
@willcumbie the health monitor popover for CLB needs it

hannahstwinter:
@willcumbie There are at least three in the CP currently - health monitoring (pictured above), monitoring alarm, and one other that is elusive (can't remember).
I see why you left it out, TBH, as there are a lot of moving parts to these controls (canonStyle, disabling, hiding, etc.) and submit and cancel are fairly predictable in this regard. In accord, relinquishing all control of these variables to the consumer feels like the most reasonable way to expose if we decide to.

test/FormPopoverSpec.jsx
hannahstwinter: It's French.

akrikos:
It does seem a pretty simple way to make an extra button available. 

I'm a bit dubious about the UX pattern of putting that button on the popover unless it's really clearly labelled. `Disable` in the given example isn't specific enough. `Disable Monitoring` could work: since the action is on the monitoring settings for the main action and the disable is for the monitoring itself, just `Disable` could be confusing. I don't know what it looks like underneath that popover to get it to show, but it could be a good idea to make that button available outside the popover...

src/FormPopover.jsx
```diff
@@ -69,6 +70,7 @@ FormPopover.propTypes = {
   // popover layout
   cancelButton: React.PropTypes.element.isRequired,
   submitButton: React.PropTypes.element.isRequired,
+  additionalControls: React.PropTypes.array,
```
jdrush89: I think this should be a node.

jdrush89:
Just one small comment, LGTM

test/FormPopoverSpec.jsx
akrikos: Too much 🇫🇷 🍷 ?

hannahstwinter:
@akrikos I can see that, but I think that really falls to the consumer to decide what their patterns for labelling should be.

test/FormPopoverSpec.jsx
wpwood: rendeurs

src/FormPopover.jsx
hannahstwinter: This was decided based on my worry regarding extending this in an inflexible fashion. I can't think of a case in which someone would want more than one additional control off of the top of my head... That said, I agree that array is vague - what about ```React.PropTypes.arrayOf(React.PropTypes.node)```?

demo/DemoPopoverSection.jsx
hannahstwinter: That does seem strange - I am seeing no warnings anywhere.

src/FormPopover.jsx
willcumbie: This was my first thought too, but it turns out that an array of nodes is still a node. https://facebook.github.io/react/docs/reusable-components.html#prop-validation

hannahstwinter: Neat - 
```
    // Anything that can be rendered: numbers, strings, elements or an array
    // (or fragment) containing these types.
    optionalNode: React.PropTypes.node
```

demo/DemoPopoverSection.jsx
hannahstwinter: Well, they're not dynamic, just a list.

akrikos:
I understand that. Other than the small nitpicks, this looks good.

PR #130, opened 2016-07-15
refactor (normalize styling): normalize use of classname module, other styling

hannahstwinter:
Including -

- reduce duplication within components by evaluating conditional display of subcomponents inline
- ternary over if/else if logic flow is simple enough
- prefer arrow function over ```function ()```
- classname module utilization - if module is being used, use consistently for all classes on an element
- curly bracket spacing (can be a lint rule, mebbe in another PR) - spaces after opening brace and before closing brace - ```{ this.is.good }``` vs ```{this.is.ew}```
- proptype declaration - use Object.keys(WHAT_NOT) whenever possible to reduce duplication while retaining concise reference to accetable values
- var declaration via let vs const - const whenever possible, using let when something will be reassigned or edited
- binding functions in constructor - this should be done to avoid binding a function more than once

If you feel that this is too many changes please let me know and I can split each point into a PR of it's own for ease. It's really all pretty straightforward, though.

### DIFF
--- a/src/Button.jsx
+++ b/src/Button.jsx
@@ -14,6 +14,12 @@ const BUTTON_TYPES = {
 };
 
 class Button extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this._handleClick = this._handleClick.bind(this);
+  }
+
   render() {
     let classes;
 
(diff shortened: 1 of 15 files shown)


adampickeral:
LGTM

PR #131, opened 2016-07-15
feat(FormField): Allow inline validation

jdrush89:
Allow users to pass an inlineValidation prop into a FormField component to make the validation be inline with the field contents.  Also allow the success prop to be a boolean or string, so that an empty message can be shown with the green check mark.

image

### DIFF
--- a/src/FormFieldValidationBlock.jsx
+++ b/src/FormFieldValidationBlock.jsx
@@ -1,15 +1,26 @@
 import React from 'react';
+import classnames from 'classnames';
 
 const FormFieldValidationBlock = (props) => {
+  const classes = classnames(
+    {
+      'rs-validation-block': !props.inline,
+      'rs-validation-inline': props.inline
+    }
+  );
+  const style = {
+    'vertical-align': 'text-top'
+  };
   return props.value ? (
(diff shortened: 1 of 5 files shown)


src/FormFieldValidationBlock.jsx
```diff
@@ -1,15 +1,26 @@
...
 
 const FormFieldValidationBlock = (props) => {
+  const classes = classnames(
+    {
+      'rs-validation-block': !props.inline,
+      'rs-validation-inline': props.inline
+    }
+  );
+  const style = {
+    'vertical-align': 'text-top'
```
hannahstwinter: Should this style be conditional? What does it do? I'm too lazy to test it myself... ;)

hannahstwinter: Also, there should be a test surrounding this.

jdrush89: It has no effect when there's a value.  When there isn't, the icon starts at the bottom of the div, this raises it up to where it would be if there was text.

jdrush89: without style: 
image

willcumbie:
:+1:

PR #132, opened 2016-07-18
feat(automation): add a script to automatically create a new release

willcumbie:
Also updated documentation. I'm leaving it in three parts deliberately, since each step requires different permissions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,12 @@ The latest stable release is available under the releases section of the reposit
 
 To build a bundled file for release run `scripts/cibuild`
 
+## Releasing
+Publishing a new release involves three steps:
+1. Create and merge a pull request which updates the changelog and version number. This may be done automatically by running the `create_release_branch.py` script.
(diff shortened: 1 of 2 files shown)


adampickeral:
do you think we could combine steps `2` and `3` into a single script?

willcumbie:
@adampickeral We could, and it would be easy, but they do require different permissions. I'm thinking of  a scenario where someone with Github write permissions runs the two scripts, and then has someone else publish to npm for them. (This would also be useful if npm were temporarily down, or someone's account were having temporary problems.)

adampickeral:
@willcumbie good points. :+1:

akrikos:
Does it feel odd to anyone else that the instructions are to publish to npm before github? I may be splitting hairs here, but I've always viewed npm as needing to mirror github even though I know they use different artifacts.

willcumbie:
Yeah, that was in the instructions as I received them, and I thought it was a bit odd. I'll switch the order in the docs.

akrikos:
Assuming it works :+1:

PR #139, opened 2016-07-27
update(documentation)

lucasrose:
**Changes Include:**
- Switched from Jekyll to GitBook
- Updated documentation to version 0.13.3
- Custom Styling
- Automatic local testing scripts
- Automatic push/commit scripts
- Documentation Instructions
- Demo: https://lucasrose.github.io/canon-react

**Relevant Files/Folders:**
- SUMMARY.md
- README.md
- documentation/
- docs/
- scripts/publish_script & scripts/update_script

**Notes:**
*- as the components are not static, they may require a page refresh for the first access.*

*- due to the way GitBook works, this will be added to the `master` branch, and leave the master branch free from generated content, the `gh-pages` branch will only be generated content*

*- the documentation can be updated by a script which will push all generated content to the `gh-pages` branch, and non-generated content to the `edit-gh-pages` branch, where both PR's can be created* https://github.com/lucasrose/canon-react/blob/8ef539ba44879da64d378d46d4c808ef101f595e/scripts/publish_gitbook

*- documentation uses gitbook-cli https://github.com/GitbookIO/gitbook-cli*

*- similar to how Redux https://github.com/reactjs/redux did their documentation http://redux.js.org/*


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+# Version History
```
willcumbie: Is this required for gitbook?

lucasrose: @willcumbie Not required, but looked nicer with a title for the page

_book/.eslintrc
```diff
@@ -0,0 +1,49 @@
+{
```
willcumbie: Are all these rules relevant here?

lucasrose: those are generated,  not entirely sure how relevant they are

adampickeral:
it looks like you're trying to merge all of this documentation into the master branch. i don't think that's the right place to be merging it

Gruntfile.js
```diff
@@ -149,6 +176,13 @@ module.exports = function (grunt) {
 
   grunt.registerTask('lint', ['eslint']);
 
+  grunt.registerTask('documentation-build', [
+    'babel:src',
```
adampickeral: do we need to build canon-react source for the documentation?

lucasrose:
@adampickeral that is how the redux docs did it, but i could try writing a script to only use gh-pages

adampickeral:
@lucasrose yeah i see that now, still looking at. don't change anything yet

adampickeral:
few thoughts:

1) should `documentation/bundle.js` be in version control since it's a generated file? (i think you could do something similar to how you add the `.nojekyll` file when publishing the docs)

2) are the `fonts/*` and `.css` files necessary? I don't see those in the master branch, which makes me think it's generated some how.

3) for the code examples themselves, we're missing the highlighting we had before. That's because we're not using any css to style it. I'm fine with not including our own css (and actually prefer it), but I wonder if there's a way to use github or gitbook's syntax highlighting within those code blocks. I guess the problem is that all of these examples are actually written in `jsx`, so it might be hard to do.

4) since there are a lot of things in `gh-pages` already for the current site, will everything just get wiped with this script, or will we need to go in and delete things manually on that branch?

lucasrose:
1. `bundle.js` is where the components are loading from so we probably will have to keep that under version control

2. fonts are needed, used only for the styling of the docs

3. I know in markdown we can use gitbook's syntax highlighting, but since the components are loaded from bundle.js and are react components, I am unsure of how to do that. The highlighting on the readme is different. I can look into applying css styling for js on those code blocks.

4. I dont think everything will be wiped with the script first. We would need to delete things manually once, then the script will wipe every commit after that.

5. Should I try to figure out a way to keep all the editing on the gh-pages branch as well?

adampickeral:
1. if you call `scripts/docbuild` at the beginning of `scripts/publish_gitbook`, then `documentation/bundle.js` will get generated, and then added later in the publish script when this happens `touch .nojekyll && git add .`

2. :white_check_mark: 
3: :white_check_mark: 
4. :white_check_mark: 

5. nah

If you can keep `bundle.js` from being stored in version control on the master branch, then I think we can publish this and think about other options for the styling separately

lucasrose:
@adampickeral so what if I just add bundle.js to the .gitignore?

adampickeral:
you don't want it to be in .gitignore because then it won't get added in the `gh-pages` branch

lucasrose:
@adampickeral i think it will get added because only the _book directory is added to the gh-pages branch

Gruntfile.js
lucasrose: @adampickeral yes, the components still have to be rendered as they were in jekyll

adampickeral: but we're using the npm module, not the raw source, right? `babel:src` transpiles the actual canon-react library, which we shouldn't need for the documentation since we use the npm module

scripts/update_gitbook
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# script/update_gitbook: build documentation site javascript then update gitbook
+
+if ! gitbook >/dev/null
+then
+        npm install gitbook-cli -g
+        gitbook install
+        npm install react-router -g
```
adampickeral: you need to break this out into a separate conditional. i also don't think react-router needs to be a global install. it can probably just get added as a dev dependency in package.json

scripts/update_gitbook
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# script/update_gitbook: build documentation site javascript then update gitbook
+
+if ! gitbook >/dev/null
+then
+        npm install gitbook-cli -g
+        gitbook install
```
adampickeral: trying to run this locally and hitting some problems. is `gitbook install` something that should only happen once, or is it ok to be run multiple times?

i'm hitting the scenario where I already have gitbook installed, but `gitbook install` hasn't been run yet when I check out this branch, so it fails for me. this is likely to be a common scenario

adampickeral: seems fine to be run multiple times, so i would move that command out of the conditional

adampickeral: let's actually make gitbook a dev-dependency too (it shouldn't need a global install either).

so with gitbook and react-router being added to package.json, this script should then call `scripts/update` first, then `gitbook install`, then build docs, then `gitbook serve`

docs/getting_started.md
```diff
@@ -0,0 +1,26 @@
+# Getting Started
+
+## Requirements
+
+Canon-React utilizes Rackspace Canon and React from Facebook, but these resources are not bundled with the Canon React library. Canon React is compatible with and requires the following bundles:
+
+**Canon Latest**
+**React v0.14.0+**
```
adampickeral: might need to make these into a list, they're getting rendered side by side in the generated html

docs/components.md
```diff
@@ -0,0 +1,16 @@
...
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+
+  <link rel="stylesheet" href="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest//canon.min.css">
+
+  <script src="https://fb.me/react-with-addons-0.14.8.min.js"></script>
+  <script src="https://fb.me/react-dom-0.14.8.min.js"></script>
+</head>
+
+<body class="rs-responsive">
```
adampickeral: i wonder if it would be better for this to just be some kind of list of (with links to) all the components in canon react, rather than sticking the documentation for each component in one file here, when we already have it broken out into multiple files

lucasrose:
@adampickeral 👍 I will make the following changes

- gitbook & react router to dev dependencies 
- `scripts/update` > `gitbook install` > `gitbook serve/publish`
- canon latest and react version # to list

docs/components.md
lucasrose: 👍

docs/edit_documentation.md
```diff
@@ -0,0 +1,88 @@
...
+2. Create the `.md` file with the path given in `SUMMARY.md`
+
+3. Add content
+
+4. Run `scripts/update_gitbook`
+
+5. Look at the changes on `localhost:4000`
+
+### Editing An Existing Page
+##### (*NOT FOR COMPONENTS*)
```
willcumbie: Instead of having this warning here, why not make the first step something like `Find the .md file (or, for component pages, the .jsx file) and make changes`?

willcumbie:
This is a really solid approach, seems good to me.

PR #143, opened 2016-07-28
style(demo): Use DetailsSection in demos, instead of divs with rs-detail-section

wpwood:
Update the demo components that display in a section to use the DetailsSection component
instead of wrapping in dovs with the rs-detail-section class. This should make the demos
update if/when the DetailsSection component is updated.


willcumbie:
:tada:

hannahstwinter:
:sheep::anchor:

lucasrose:
🚢 🚀

PR #145, opened 2016-08-02
fix(List Table): Fix incorrect direction and overlay status proptype declarations

wpwood:
SortDirection and OverlayStatus are mappings of key names (SortDirection.ASCENDING, frex) to values ('asc') so that a developer can use the key names to pass the values. Therefore, the PropTypes declaration should be validating on the value that would be passed, not the key name.


willcumbie:
:+1:

j00k:
:rocket:

PR #146, opened 2016-08-03
feat(new documentation)

lucasrose:
Switched from Jekyll to Gitbook

https://lucasrose.github.io/canon-react

closes #139


lucasrose:
we will also have to merge the`gh-pages` branch

adampickeral:
`.nojekyll` doesn't need to be checked in

adampickeral:
`_book` shouldn't be checked in either, right?

lucasrose:
@adampickeral removing those now

scripts/publish_gitbook
```diff
@@ -0,0 +1,19 @@
...
+gitbook build
+ORIGIN=$(git remote get-url origin)
+cd _book
+git init
+git remote -v
+git commit --allow-empty -m 'update(documentation)'
+git checkout -b gh-pages
+touch .nojekyll && git add . && git add ../documentation/bundle.js
+git commit --amend --no-edit
+git push $ORIGIN gh-pages --force
```
adampickeral: i assume this actually updates gh-pages, but what are the following lines for?

lucasrose: @adampickeral 
we cannot add bundle.js to the .gitignore or the gh-pages branch will not build properly, so we remove bundle.js after then update the edit documentation branch.

If someone makes changes the process will look like:
1. Make change
2. Test locally with `scripts/update_gitbook` & etc.
3. Run `scripts/publish_gitbook`
4. Create PR for `edit-documentation` branch to be merged in master, then create PR for `gh-pages` to be merged into gh-pages.

lucasrose: And how the repo works, we cannot make changes without a PR, so we need a PR for the `gh-pages` branch as well

adampickeral: if i haven't forked though, the force push to gh-pages is going to update the live site, right?

i think the workflow for docs changes will be a little different than source changes. people could still fork and push things into master, but to actually publish to gh-pages, that can (and should) be done from upstream, without the need for a PR

so i think the workflow would be more like:
1. Make a docs change (in fork or in the main repo)
2. Test locally
3. Merge that PR
4. Checkout and pull in upstream master.
5. Run `scripts/publish_gitbook`
6. :moneybag: 

So I don't think we need the separate step to push something up to an `edit-documentation` branch.

PR #148, opened 2016-08-12
Hack day/multi step progress bar

zghaed:
Added MultiStepProgressItem and MultiStepProgressBar components
and their tests and demos to canon-react.

screen shot 2016-08-12 at 1 42 39 pm

### DIFF
--- a/src/MultiStepProgressBar.jsx
+++ b/src/MultiStepProgressBar.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const SIZE_CLASSES = {
+  'xsmall': 'rs-progress-xsmall',
+  'small': 'rs-progress-small',
+  'medium': 'rs-progress-medium',
+  'large': 'rs-progress-large',
+  'xlarge': 'rs-progress-xlarge'
+};
+
+class MultiStepProgressBar extends React.Component {
+  render() {
+    const sizeClass = classNames(
+      'rs-progress',
+      SIZE_CLASSES[this.props.size]
+    );
+
+    return (
+      <div className={ sizeClass }>
+        <div className="rs-progress-inner">
+          {this.props.children}
+        </div>
+      </div>
+    );
+  }
+}
+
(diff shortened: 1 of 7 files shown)


src/MultiStepProgressBar.jsx
```diff
@@ -0,0 +1,33 @@
...
+  render() {
+    const sizeClass = classNames(
+      'rs-progress',
+      SIZE_CLASSES[this.props.size]
+    );
+
+    return (
+      <div className={ sizeClass }>
+        <div className="rs-progress-inner">
+          {this.props.children}
```
hannahstwinter: spaces - ```{ this.props.children }```

src/MultiStepProgressItem.jsx
```diff
@@ -0,0 +1,48 @@
...
+  render() {
+    const statusClass = classNames(
+      'rs-bar',
+      STATUS_CLASSES[this.props.status],
+      TYPE_CLASSES[this.props.type]
+    );
+    const style = { 'width': `${ this.props.progress }%` };
+
+    return (
+      <div className="rs-segment" style={ style }>
```
hannahstwinter: Had you tried using the existing ProgressBar component here? I'm curious as to whether that would work... There's a fair amount of duplication here...

hannahstwinter: If it's not reasonable we should at least pull out the shared class maps.

zghaed: Yeah, I know there's a fair amount of duplication here. MultiStepProgressItem has `caption` which ProgressBar doesn't.

zghaed: @hannahstwinter I can add an optional caption to ProgressBar and remove this component. What do you think?

jdrush89:
Check out the commit guidelines https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines
You should do a rebase -i and squash to one commit, something like:
feat: adding MultiStepProgressItem and MultiStepProgressBar

PR #152, opened 2016-08-30
fix(ListTable) explicit React import

draganaperez:


adampickeral:
Add it to these components too:
https://github.com/rackerlabs/canon-react/blob/master/src/CheckboxColumnHeader.jsx
https://github.com/rackerlabs/canon-react/blob/master/src/EmbeddedListTable.jsx
https://github.com/rackerlabs/canon-react/blob/master/src/EmptyOverlay.jsx
https://github.com/rackerlabs/canon-react/blob/master/src/ErrorOverlay.jsx
https://github.com/rackerlabs/canon-react/blob/master/src/StatusColumnHeader.jsx
https://github.com/rackerlabs/canon-react/blob/master/src/TextColumnHeader.jsx

PR #154, opened 2016-08-31
fix(automation): add tar zip step to release script

willcumbie:
This resolves a few bugs with the automation script that were due to the idiosyncratic setup I had locally. We don't need `sudo` to run the `cibuild` script, and since tar-ing the minified JS doesn't happen by default, that needs to happen within this script.


adampickeral:
:shipit:

PR #155, opened 2016-09-22
feat(TooltipTrigger): Add onDisplay hook prop

willcumbie:
Add a prop "onDisplay" which will be fired when the tooltip is shown, which works as a seam for adding tracking to tooltips.

### DIFF
--- a/test/TooltipTriggerSpec.jsx
+++ b/test/TooltipTriggerSpec.jsx
@@ -4,14 +4,15 @@ import TooltipTrigger from '../transpiled/TooltipTrigger';
 import TestUtils from 'react-addons-test-utils';
 
 describe('TooltipTrigger', () => {
-  let trigger, tether, triggerParent;
+  let trigger, tether, triggerParent, onShowSpy;
 
-  const renderTrigger = (placement = 'bottom-right') => {
+  onShowSpy = jasmine.createSpy('onShow');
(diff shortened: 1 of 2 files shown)


adampickeral:
looks fantastic to me

test/TooltipTriggerSpec.jsx
```diff
@@ -4,14 +4,15 @@ import TooltipTrigger from '../transpiled/TooltipTrigger';
...
+  let trigger, tether, triggerParent, onDisplay;
 
   const renderTrigger = (placement = 'bottom-right') => {
+    onDisplay = jasmine.createSpy('onDisplay');
     tether = jasmine.createSpyObj('tether', ['destroy']);
     spyOn(TooltipTrigger.prototype, '_createTether').and.returnValue(tether);
 
     trigger = TestUtils.renderIntoDocument(
-      <TooltipTrigger placement={placement} content={'The tooltip content'}>
+      <TooltipTrigger placement={placement} content={'The tooltip content'} onDisplay={onDisplay}>
```
wpwood: Do we need a test for when onDisplay is not supplied? I could go either way on that one.

hannahstwinter:
LGTM!

test/TooltipTriggerSpec.jsx
willcumbie: You're right, that was an oversight on my part. Added.

PR #158, opened 2016-09-26
fix(Embedded List Table): Provide default overlay components

jdrush89:
The List Table provides default overlay components, but the embedded list table did not.  Adding them here as well.


adampickeral:
LGTM

PR #159, opened 2016-09-30
fix(DropdownItem): pass innerProps to innerElement

j00k:
otherwise you can't have dropdownitems that actually have links :relaxed:


adampickeral:
:+1:

PR #162, opened 2016-10-07
Add multistep progress components

jdrush89:
I extracted some of the common functionality of the existing progress bar and multi step progress bar components into a ProgressSegment and ProgressBarContainer components.  The normal ProgressBar now uses these components and works the same, while multistep progress bars can be created by putting multiple segments into a container.
image


Jigarsolanki:
LGTM

willcumbie:
:+1: Don't forget to squash the commits into one before merging.

PR #163, opened 2016-10-12
Add optional prop to display additional content on the FormPopover footer

ilmotta:
This PR adds a new prop to the `FormPopover` component in order to allow developers to display additional right-aligned with the control buttons/links (e.g. submit and cancel).

#### Problem

I have a submittable popover component using the `FormPopover` component and I need to display a right-aligned tooltip on the footer. There is no appropriate way to use a `FormPopover` and display additional content after the control buttons. The only thing similar is the `additionalControls` prop, but it is rendered before the cancel component.

#### Updates based on suggestions

1. Renamed the new prop from `additionalInfo` to `additionalFooterContent`
2. Always right-align the extra content (see screenshots below). The implementation is straightforward: wrap the content in a `<div />` with class `rs-pull-right` (tests updated accordingly).
3. Improved demo

#### Tasks

- [x] Add support for new prop in `FormPopover`
- [x] Implement demo
- [x] Update changelog
- [x] Always right-align the additional content

---

### Screenshot: submittable popover displaying additional content as a `<TooltipTrigger />`

image

### Screenshot: The extra content plays nicely with the processing indicator on submittable popovers

image

### DIFF
--- a/demo/DemoPopoverSection.jsx
+++ b/demo/DemoPopoverSection.jsx
@@ -5,6 +5,8 @@ import Button from './Button';
 import DemoFormPopover from './DemoFormPopover';
 import DemoPopover from './DemoPopover';
 import DetailsSection from './DetailsSection';
+import StatusIndicator from './StatusIndicator';
+import TooltipTrigger from './TooltipTrigger';
 
 class DemoPopoverSection extends React.Component {
   constructor(props) {
@@ -97,6 +99,32 @@ class DemoPopoverSection extends React.Component {
                   onSubmit={ () => { this.setState({ formPopoverSubmitting: true }) } }
                   onRequestClose={ () => { this.setState({ formPopoverOpen: false }) } }
                   additionalControls={ [ <Button onClick={ (e) => { e.preventDefault(); } } canonStyle="secondary">Opt. Addl Controls</Button> ] } />
+              </td>
+            </tr>
+            <tr>
+              <td>
+                <Button
+                  id="form-popover-with-additional-footer-content-button"
+                  onClick={ () => this.setState({
+                    formPopoverWithAdditionalFooterContentOpen: true,
+                    formPopoverWithAdditionalFooterContentSubmitting: false
+                  }) }>
+                  Submittable with additional footer content
+                </Button>
+                <DemoFormPopover
+                  placement="right"
+                  processing={ this.state.formPopoverWithAdditionalFooterContentSubmitting }
+                  target={ () => document.getElementById('form-popover-with-additional-footer-content-button') }
+                  isOpen={ !!this.state.formPopoverWithAdditionalFooterContentOpen }
+                  onSubmit={ () => this.setState({ formPopoverWithAdditionalFooterContentSubmitting: true }) }
+                  onRequestClose={ () => this.setState({ formPopoverWithAdditionalFooterContentOpen: false }) }
+                  additionalFooterContent={
+                    <TooltipTrigger
+                      placement="bottom"
+                      content="In this example, the additional footer content is a TooltipTrigger, but it could be any React node/component or even plain text.">
+                      <StatusIndicator>Tooltip</StatusIndicator>
+                    </TooltipTrigger>
+                  } />
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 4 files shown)


hannahstwinter:
A few things I am wondering about - the demo seems to be very contrived and offers little insight into why you would use this feature. Do you think that passing in a tooltip trigger with the tooltip itself indicating 'Optional additionalControlsContent' would be any clearer? Also, maybe ```additionalControlsContent``` for the name, as it indicates where the additional info is going to go. Lastly, is this a common enough pattern to warrant putting it into Canon-React? I haven't seen tooltips in footers anywhere... I'm not 100% opposed to it, I just worry about prematurely optimizing.

adampickeral:
we often have pricing tooltips in footers, don't we?

hannahstwinter:
@adampickeral Snap, that's right! Those are typically right aligned, yeah?

ilmotta:
@hannahstwinter, thanks for sharing your thoughts.

I agree the demo can be improved. I went on the same route as the other `additionalControls` prop and tried to keep them similar. Any suggestions?

> Do you think that passing in a tooltip trigger with the tooltip itself indicating 'Optional additionalControlsContent' would be any clearer?

My problem is strictly related to displaying tooltips on the footer, but I tried to keep names less specific. I think your suggestion of `additionalControlsContent` is a good one!

> is this a common enough pattern to warrant putting it into Canon-React?

I found a couple places with pricing tooltips as Adam said. One of them even duplicate the same behaviour as the FormPopover and implements a `tooltip` prop. My motivation was to keep things simple and perhaps reuse this feature in the future for refactorings or just for new popovers with pricing tooltips.

hannahstwinter:
I can definitely see having this in Canon-React - for some reason I had forgotten entirely about pricing tooltips. That said, as I mentioned above, I'm pretty sure those are right aligned - is that something that we would want to enforce here? If not, there would need to be a custom rule/custom style on the element every time it is used for that application. I wonder about exposing that as a toggle-able feature... That would be too much, yeah? I'm not sure what the best approach on that one is...
I agree that the demo could accidentally imply that the extra content should be a tooltip... What about tooltip underlined text - <u>Opt. Addl Controls Content</u> - and have the tooltip explain that the extra content can be anything? Yay for using existing features to make your demos more robust, eh? 😄

ilmotta:
> I'm pretty sure those are right aligned - is that something that we would want to enforce here?

I kept asking myself if the popover should be more helpful for the "90%" use case (i.e. display something right aligned). You're right, the application would need at least a global CSS rule to push that extra content to the right. It's flexible, but not so convenient. btw, Flexbox would be handy here 😄. I'm not sure if Canon has recommendations for this kind of trade-off; that'd help.

I'm going to implement your suggestion for the demo with the tooltip.

willcumbie:
I feel like `additionalInfo` is a bit of a strange name. Could it be something like `additionalFooterContent`?

demo/DemoPopoverSection.jsx
```diff
@@ -101,6 +103,32 @@ class DemoPopoverSection extends React.Component {
             </tr>
             <tr>
               <td>
+                <Button
+                  id="form-popover-with-additional-footer-content-button"
+                  onClick={ () => this.setState({
+                    formPopoverWithAdditionalFooterContentOpen: true,
+                    formPopoverWithAdditionalFooterContentSubmitting: false
+                  }) }>
+                  Submittable with additional footer content
```
hannahstwinter: The separate demo is a good way to keep things clean. :+1:

adampickeral:
once you have finished with code changes, please squash everything into a single commit.

also remove the changes to the changelog, that happens during the release process

ilmotta:
I'll make the changes @adampickeral. btw, I just searched in the repo and docs and there is no clear mention or explanation for the squash practice but I've seen a reminder on almost every PR. Is this something we should improve on the Contributing documentation (not in this PR of course)?

I found this bullet-point in the docs:

> Intermediate and work-in-progress commits should be rebased out

adampickeral:
we can update the contributing guidelines. i do consider all the commits in here to be intermediate ones though that are either addressing code review changes or are really just parts of the whole feature

willcumbie:
LGTM

PR #167, opened 2017-02-28
feat(FormField): render dom attributes

adampickeral:
add passed in dom attributes to the rendered control goup in form fields

### DIFF
--- a/src/FormField.jsx
+++ b/src/FormField.jsx
@@ -4,21 +4,22 @@ import FormFieldValidationBlock from './FormFieldValidationBlock';
 import classnames from 'classnames';
 
 const FormField = (props) => {
+  const { error, success, label, children, help, inlineValidation, ...rest } = props;
   const classes = classnames(
     'rs-control-group',
     {
-      'error': props.error,
-      'success': props.success
+      'error': error,
+      'success': success
     }
   );
 
   return (
-    <div className={ classes }>
-      <label className="rs-control-label">{ props.label }</label>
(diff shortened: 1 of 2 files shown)


j00k:
:ship:

willcumbie:
Yikes! Good fix.

src/FormField.jsx
```diff
@@ -4,21 +4,22 @@ import FormFieldValidationBlock from './FormFieldValidationBlock';
 import classnames from 'classnames';
 
 const FormField = (props) => {
+  const { error, success, label, children, help, inlineValidation, ...rest } = props;
```
hannahstwinter: :+1:

src/FormField.jsx
```diff
@@ -4,21 +4,22 @@ import FormFieldValidationBlock from './FormFieldValidationBlock';
...
-      'success': props.success
+      'error': error,
+      'success': success
     }
   );
 
   return (
-    <div className={ classes }>
-      <label className="rs-control-label">{ props.label }</label>
+    <div {...rest} className={ classes }>
```
MACSkeptic: would you want to also append user provided classes to the `className` here?

in other words, should `<FormField className="x" ...` include `x` + the default class names?

adampickeral: yes, yes you would. i paid no attention to the code above that was using `classnames` and assumed it was already doing that

adampickeral: addressed in #169

PR #171, opened 2017-06-14
Upgrading to latest react

austburn:


src/Button.jsx
```diff
@@ -22,24 +23,25 @@ class Button extends React.Component {
 
   render() {
     let classes;
+    const { canonStyle, enabled, ...rest } = this.props;
```
austburn: would love some feedback on `...rest`

src/FormFieldValidationBlock.jsx
```diff
@@ -19,8 +20,8 @@ const FormFieldValidationBlock = (props) => {
 };
 
 FormFieldValidationBlock.propTypes = {
-  value: React.PropTypes.node,
-  inline: React.PropTypes.bool
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
austburn: this was to meet requirements specified in the test https://github.com/rackerlabs/canon-react/pull/171/files#diff-863bbaf5f87cf22e969f1461c5f395c3R33

test/FormPopoverSpec.jsx
```diff
@@ -91,21 +92,26 @@ describe('FormPopover', () => {
...
-      [ popoverBody, popoverFooter ] = form.props.children;
-      [
-        errorIndicator,
-        submit,
-        additionalControls,
-        cancel,
-        additionalFooterContent,
-        processingIndicator
-      ] = popoverFooter.props.children;
+      let submitIndex, footerComponents, innerComponents;
```
austburn: Without a doubt the least straightforward of the test changes, most of this stems from the optional `additionalControls`/`additionalFooterContent` `props`. I played around with a couple different ways of making the decisions appear explicit... but with the smack dab in the middle of the `PopoverFooter`, things got a bit wonky.

austburn: I guess the biggest takeaway is that the tests for the most part didn't change and in some cases actually improved imo - see https://github.com/rackerlabs/canon-react/pull/171/files#diff-1ba8c3e7790dd15e89c0754cd374e08dR205.

test/FormSpec.jsx
```diff
@@ -1,75 +1,59 @@
...
   });
 
   it('adds a valid size class', () => {
-    renderWithProps({ size: 'xsmall', horizontal: false });
-    const form = renderer.getRenderOutput();
-
-    expect(form.props.className).toEqual('rs-form-xsmall');
-  });
-
-  it('does not add an invalid size classes', () => {
```
austburn: removed this test as it was basically testing `PropTypes`

src/Button.jsx
adampickeral: i approve of this method

demo/index.html
```diff
@@ -3,8 +3,8 @@
   <head>
     <link rel="stylesheet" type="text/css" href="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.css">
     <script type="application/javascript" src="https://eafdbc63c97ce6bec9ef-b0a668e5876bef6fe25684caf71db405.ssl.cf1.rackcdn.com/v1-latest/canon.min.js"></script>
-    <script src="https://fb.me/react-0.14.8.min.js"></script>
-    <script src="https://fb.me/react-dom-0.14.8.min.js"></script>
+    <script src="https://fb.me/react-15.5.4.min.js"></script>
```
adampickeral: it's already out of date! :trollface:

src/Button.jsx
adampickeral: you should also pull `className` and `hidden` out of here since they're being used in the `classes` snippet below

src/DetailsSection.jsx
```diff
@@ -18,7 +19,10 @@ class DetailsSection extends React.Component {
   }
 
   render() {
-    let { collapsible, isLoading } = this.props, { isCollapsed } = this.state;
+    let divProps = Object.assign({}, this.props), { isCollapsed } = this.state;
+    delete divProps.initialCollapsed;
+    delete divProps.onToggleCollapsed;
+    let { collapsible, isLoading, subtitle, ...rest } = divProps;
```
adampickeral: also need to get `title` and `children`

src/Button.jsx
```diff
@@ -23,26 +23,27 @@ class Button extends React.Component {
 
   render() {
     let classes;
-    const { canonStyle, enabled, ...rest } = this.props;
+    const { canonStyle, enabled, className, hidden, children, ...rest } = this.props;
+    delete rest.onClick;
```
adampickeral: you shouldn't need to delete this from rest because it'll get overwritten by the onClick definition on the element

src/DropdownItem.jsx
```diff
@@ -17,24 +17,23 @@ class DropdownItem extends React.Component {
   }
 
   render() {
-    const liProps = Object.assign({}, this.props);
-    delete liProps.hideCallback;
-    delete liProps.innerProps;
-    delete liProps.enabled;
+    const { enabled, innerProps, type, className, ...rest } = this.props;
+    delete rest.onClick;
```
adampickeral: same here, no need to delete onClick

Gruntfile.js
```diff
@@ -89,7 +89,7 @@ module.exports = function (grunt) {
           'test_bundle.js': ['test-built/**/*.js']
         },
         options: {
-          external: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment'],
+          external: ['react/lib/ReactContext', 'react/lib/ExecutionEnvironment', 'react/addons', 'react-addons-test-utils'],
```
ashimaathri: What are the addons for?

src/Button.jsx
```diff
@@ -22,25 +23,26 @@ class Button extends React.Component {
...
+        <button { ...rest } className={ classes } onClick={ this._handleClick.bind(this) }>
+          <span className='rs-cog'></span> { children } <span className='rs-caret'></span>
         </button>
       );
     }
 
     return (
-      <button {...this.props} className={classes} onClick={this._handleClick.bind(this)}>
-         {this.props.children}
+      <button { ...rest } className={ classes } onClick={ this._handleClick.bind(this) }>
```
ashimaathri: How did you know that not all of the attributes of the props object was used in a Button?

Gruntfile.js
austburn: So, some of the dependencies of `canon-react` have logic to support multiple versions of `react` over time, and they tend to perform conditional imports. This just tells `browserify` when you see and import with something in `external` (which are deprecated for the current version of `react`), don't worry about bundling them.

src/Button.jsx
ashimaathri: It looks like Button does take `canonStyle`, `enabled` and `hidden` which are not in `rest` right?

ashimaathri: Oh but this is a `button` not a `Button`

austburn: For most of these, I used the `propTypes` at the bottom of the file to determine what I pull out.

ashimaathri: Just checked https://facebook.github.io/react-native/docs/button.html#props. Looks good :+1:

src/DetailsSection.jsx
```diff
@@ -18,11 +19,22 @@ class DetailsSection extends React.Component {
...
+      title,
+      headers,
+      isLoading,
+      collapsible,
+      className,
+      subtitle,
+      children,
+      ...rest } = this.props;
+    delete rest.initialCollapsed;
+    delete rest.onToggleCollapsed;
```
ashimaathri: Why are these two props passed if they aren't used?

austburn: They're referenced above, but should not be passed to the `div` below.

src/DropdownItem.jsx
```diff
@@ -16,19 +17,22 @@ class DropdownItem extends React.Component {
   }
 
   render() {
+    const { enabled, innerProps, type, className, ...rest } = this.props;
+    delete rest.hideCallback;
```
ashimaathri: Same question for unused prop here

src/DetailsSection.jsx
ashimaathri: Would it be possible to pull them out from rest in the destructing step itself like
`{title, headers, initialCollapsed, ...rest} = this.props` or would the linter complain about unused variables?

ashimaathri: Just curious, no strong feelings either way

